PR #3, opened 2015-01-05
getting current user with whoami like function rather than from env variable

stibi:
Hi guys,
I've just had quite a problem with i3lock, I was not able to unlock…I was entering the correct password, that's for sure. The PAM configuration seems to be correct too.

The problem was with the `$USER` environmental value…see:

    $ echo $USER
    %{%{%}%}%n

That's not my username :) And because i3lock gets the username from that env value...it fails to unlock against wrong/non-existing user.

    ...
    if ((username = getenv("USER")) == NULL)
    ...

It's obvious I should first fix the wrong content of my `$USER`, and I'll do it, but I was also thinking that it might be a good idea to change how i3lock find out who is the current user of a session.
After a quick googling, I've found that it is easy to write own "whoami" function, so I did it and here is the pull request.

What do you think?
Btw, this is my first shot with C programming, although it's almost only copy-pasta job, maybe I did something wrong, I have to learn more for example about pointers :)
If you have any suggestion, I'll be happy to fix any problem there could be.

Thanks!

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -8,6 +8,7 @@
  */
 #include <stdio.h>
 #include <stdlib.h>
+#include <pwd.h>
 #include <string.h>
 #include <unistd.h>
 #include <stdbool.h>
@@ -658,8 +659,18 @@ static void raise_loop(xcb_window_t window) {
     }
 }
 
+char* whoami(void) {
+    uid_t uid = geteuid();
+    struct passwd *pw = getpwuid(uid);
+    if (pw) {
+        return pw->pw_name;
+     } else {
+        errx(EXIT_FAILURE, "Username not known!\n");
+     }
+}
+
 int main(int argc, char *argv[]) {
-    char *username;
+    char *username = whoami();
     char *image_path = NULL;
     int ret;
     struct pam_conv conv = {conv_callback, NULL};
@@ -683,9 +694,6 @@ int main(int argc, char *argv[]) {
         {"show-failed-attempts", no_argument, NULL, 'f'},
         {NULL, no_argument, NULL, 0}
     };
-
-    if ((username = getenv("USER")) == NULL)
-        errx(EXIT_FAILURE, "USER environment variable not set, please set it.\n");
 
     char *optstring = "hvnbdc:p:ui:teI:f";
     while ((o = getopt_long(argc, argv, optstring, longopts, &optind)) != -1) {
--- a/i3lock.c
+++ b/i3lock.c
@@ -9,6 +9,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <pwd.h>
+#include <sys/types.h>
 #include <string.h>
 #include <unistd.h>
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


i3lock.c
```diff
@@ -8,6 +8,7 @@
  */
 #include <stdio.h>
 #include <stdlib.h>
+#include <pwd.h>
```
stapelberg: According to the manpage getpwnam(3), you must also include sys/types.h

i3lock.c
```diff
@@ -658,8 +659,18 @@ static void raise_loop(xcb_window_t window) {
     }
 }
 
+char* whoami(void) {
+    uid_t uid = geteuid();
+    struct passwd *pw = getpwuid(uid);
+    if (pw) {
+        return pw->pw_name;
+     } else {
+        errx(EXIT_FAILURE, "Username not known!\n");
```
stapelberg: Make this `err(EXIT_FAILURE, "getpwuid() failed");`. See the manpage err(3) — `err()` will describe `errno`, which `getpwuid()` sets on failure.

i3lock.c
```diff
@@ -658,8 +659,18 @@ static void raise_loop(xcb_window_t window) {
     }
 }
 
+char* whoami(void) {
+    uid_t uid = geteuid();
+    struct passwd *pw = getpwuid(uid);
+    if (pw) {
```
stapelberg: nit: in the i3 codebase, we prefer `if (pw != NULL) {`

i3lock.c
```diff
@@ -658,8 +659,18 @@ static void raise_loop(xcb_window_t window) {
     }
 }
 
+char* whoami(void) {
```
stapelberg: nit: in the i3 codebase, we prefer `char *whoami(void) {`

i3lock.c
```diff
@@ -684,9 +695,6 @@ int main(int argc, char *argv[]) {
         {NULL, no_argument, NULL, 0}
     };
 
-    if ((username = getenv("USER")) == NULL)
```
stapelberg: It’s slightly cleaner to handle errors in the calling program instead of having a function which then terminates the program (see also: libraries must never call `exit()`).

In order to make this pull request smaller (which is always good), you could just call `getpwuid(geteuid())` in the main function and not introduce a new function at all.

i3lock.c
```diff
@@ -658,8 +659,18 @@ static void raise_loop(xcb_window_t window) {
     }
 }
 
+char* whoami(void) {
+    uid_t uid = geteuid();
```
stapelberg: See http://docstore.mik.ua/orelly/networking/puis/ch04_03.htm — I think we should use `getuid()` instead of `geteuid()` here.

i3lock.c
stibi: good, that sounds like good idea, pushing the changes in few minutes...

luislobo:
In my case, I had to use `sudo chmod +r /etc/shadow` to make it work under Ubuntu with shadow password

MDIB:
@luislobo  men this tip save me, my i3lock just stop unlocking my screen a week ago and your `sudo chmod +r /etc/shadow` script was the only fix I could find, BTW I am on an arch distro

PR #5, opened 2015-02-11
Add capslock indicator option

boeglin:
Give visual feedback on the state of the caps lock mod.
Disabled by default.

Signed-off-by: Alexandre Boeglin <alex@boeglin.org>

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -59,6 +59,8 @@ static bool beep = false;
 bool debug_mode = false;
 static bool dpms = false;
 bool unlock_indicator = true;
+bool capslock_active = false;
+bool numlock_active = false;
 static bool dont_fork = false;
 struct ev_loop *main_loop;
 static struct ev_timer *clear_pam_wrong_timeout;
@@ -269,6 +271,11 @@ static void input_done(void) {
     if (debug_mode)
         fprintf(stderr, "Authentication failure\n");
 
+    /* Get state of Caps and Num lock modifiers, to be displayed in
+     * STATE_PAM_WRONG state */
+    capslock_active = xkb_state_mod_name_is_active(xkb_state, XKB_MOD_NAME_CAPS, XKB_STATE_MODS_EFFECTIVE);
+    numlock_active = xkb_state_mod_name_is_active(xkb_state, XKB_MOD_NAME_NUM, XKB_STATE_MODS_EFFECTIVE);
+
     pam_state = STATE_PAM_WRONG;
     failed_attempts += 1;
     clear_input();
@@ -319,7 +326,7 @@ static void handle_key_press(xcb_key_press_event_t *event) {
     bool composed = false;
 
     ksym = xkb_state_key_get_one_sym(xkb_state, event->detail);
-    ctrl = xkb_state_mod_name_is_active(xkb_state, "Control", XKB_STATE_MODS_DEPRESSED);
+    ctrl = xkb_state_mod_name_is_active(xkb_state, XKB_MOD_NAME_CTRL, XKB_STATE_MODS_DEPRESSED);
 
     /* The buffer will be null-terminated, so n >= 2 for 1 actual character. */
     memset(buffer, '\0', sizeof(buffer));
--- a/unlock_indicator.c
+++ b/unlock_indicator.c
@@ -9,6 +9,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <string.h>
 #include <math.h>
 #include <xcb/xcb.h>
 #include <ev.h>
@@ -43,6 +44,10 @@ extern uint32_t last_resolution[2];
 
 /* Whether the unlock indicator is enabled (defaults to true). */
 extern bool unlock_indicator;
+
+/* Whether the capslock and numlock mods are active or not. */
+extern bool capslock_active;
+extern bool numlock_active;
 
 /* A Cairo surface containing the specified image (-i), if any. */
 extern cairo_surface_t *img;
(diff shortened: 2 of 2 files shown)


i3lock.1
```diff
@@ -85,6 +86,11 @@ show you the current PAM state (whether your password is currently being
 verified or whether it is wrong).
 
 .TP
+.B \-C, \-\-capslock-indicator
+Enable the capslock indicator. This will give visual feedback on the state of
+the caps lock mod.
```
stapelberg: Please improve the docs: describe how the visual indicator looks like, i.e. how the user can tell if caps lock is on. Also describe the motivation behind it, i.e. that the computer cannot be unlocked while caps lock is on, unless you are aware of it.

i3lock.c
```diff
@@ -348,27 +353,29 @@ static void handle_key_press(xcb_key_press_event_t *event) {
         return;
     }
 
-    if ((input_position + 8) >= sizeof(password))
-        return;
+    if (!(capslock_indicator && ksym == XKB_KEY_Caps_Lock)) {
```
stapelberg: Are you trying to ignore XCB_KEY_Caps_Lock? If so, add it to the switch/case statement above, that will make the diff much cleaner.

i3lock.c
```diff
@@ -58,6 +58,8 @@ static bool beep = false;
 bool debug_mode = false;
 static bool dpms = false;
 bool unlock_indicator = true;
+static bool capslock_indicator = false;
```
stapelberg: Why is this disabled by default? I think it makes more sense to enable it by default.

unlock_indicator.c
```diff
@@ -21,8 +21,8 @@
 #include "xinerama.h"
 
 #define BUTTON_RADIUS 90
-#define BUTTON_SPACE (BUTTON_RADIUS + 5)
-#define BUTTON_CENTER (BUTTON_RADIUS + 5)
+#define BUTTON_SPACE (BUTTON_RADIUS + 20)
```
stapelberg: I’m not convinced these visual changes are clear enough/make sense. Can you attach a screenshot of how this looks like with/without caps lock?

stapelberg:
On a higher level, why only caps lock and not also other lock states, e.g. numlock?

boeglin:
Hi, I just pushed an ammended commit:
- doc is more verbose, as requested
- I moved the "XCB_KEY_Caps_Lock" handling in the switch/case

A screenshot of the unlock indicator, while caps lock is enabled and unlock_state equals STATE_KEY_ACTIVE, is available at https://boeglin.org/~alex/i3lock_capslock_indicator.png

I disabled it by default, since it might leak info to an eavesdropper looking at the screen while the password is being typed.

I only handled the caps lock case, since it seems to me that it happens much more frequently than the num lock case.

Best regards,
Alex

stapelberg:
> A screenshot of the unlock indicator, while caps lock is enabled and unlock_state equals STATE_KEY_ACTIVE, is available at https://boeglin.org/~alex/i3lock_capslock_indicator.png

I don’t think it’s intuitively clear what a thicker border means. Why not display “[caps|num|caps&num]lock” where we otherwise display “wrong”?

> I disabled it by default, since it might leak info (caps lock status) to an eavesdropper looking at the screen while the password is being typed.

A better way to handle this concern is by only displaying the indicator when no keys (except for lock keys) were pressed so far. That way, the user will see that the lock key is active after entering the password wrong, and users who actually use lock keys while entering their password will not even notice.

> I only handled the caps lock case, since it seems to me that it happens much more frequently than the num lock case.

That’s likely true, but it doesn’t seem to me like also handling numlock is much more effort, so please extend the patch to handle any kind of lock key.

boeglin:
Ok, here's a new patch, that simply displays "Caps, Num locked" below "wrong!" when appropriate, using a slightly smaller font.

Alex

unlock_indicator.c
```diff
@@ -229,6 +234,30 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
             cairo_close_path(ctx);
         }
 
+        if (pam_state == STATE_PAM_WRONG && (capslock_active || numlock_active)) {
+            cairo_text_extents_t extents;
+            double x, y;
+            char lock_text[32] = "";
```
stapelberg: It seems like a nitpick at first, but considering i3lock’s security-sensitive nature, I’d like to avoid statically sized buffers and `strcat()`, so please change this to something like:

```c
char *lock_text = NULL;
if (asprintf(&lock_text, "%s%s%s locked",
    (capslock_active ? "Caps" : ""),
    (capslock_active && numlock_active ? ", " : ""),
    (numlock_active ? "Num" : "")) != -1) {
  // cairo code…
  free(lock_text);
}
```

boeglin:
Done

Merovius:
Not to be itty-nitty-picky, but there are a lot more locks than just caps- and num. There's also shift-lock and e.g. on the neo-layout two to four other type of locks (I think you can lock into every layer of which there are six in total. At least layers two to four are possible, that happens to me all the time). Don't know if you can even catch every kind of lock there is, lest describe it. And I also personally don't care a lot about this feature, so it won't bother me. Just wanted to point that out.

stapelberg:
> but there are a lot more locks than just caps- and num

I think people interested in i3lock displaying those should send a separate pull request. I’ve spent a couple of minutes looking into this and it’s not obvious.

stapelberg:
>> but there are a lot more locks than just caps- and num

> I think people interested in i3lock displaying those should send a separate pull request. I’ve spent a couple of minutes looking into this and it’s not obvious.

So, I’ve found some more time to play around with this, and I think it would indeed be better if we just looped through all the modifiers and added their names to a list of active locks. That trivially covers neo’s shift lock, and as a bonus all the other, weirder locks, too :).

Here’s some example code to get this going:

```c
    DEBUG("num_mods = %d\n", xkb_keymap_num_mods(xkb_keymap));
    for (xkb_mod_index_t idx = 0;
            idx < xkb_keymap_num_mods(xkb_keymap);
            idx++) {
        DEBUG("checking whether mod idx = %d (name = %s) is active: %d\n",
                idx, xkb_keymap_mod_get_name(xkb_keymap, idx),
                xkb_state_mod_index_is_active(xkb_state, idx, XKB_STATE_MODS_EFFECTIVE));
    }   
```

@boeglin Could you please change your pull request to display all modifiers like outlined above? I think that will make your feature much more useful for users of all sorts of layouts (our user base is pretty international). Thanks a lot for your patience on this.

Deiz:
Took a quick crack at this in Deiz/i3lock@4bb15b45 (based on this pull request).

I opted to manually replace the names of capslock ("Lock" in xkb), and numlock ("Mod2" in xkb) to avoid undermining the initial purpose of this PR.

My one concern is that the button can only comfortably fit 2-3 modifiers, though perhaps that's sufficient.

stapelberg:
@Deiz Can you send that as a pull request so that I can review it?

stapelberg:
This is obsolete with #10 having been merged.

PR #6, opened 2015-02-11
Add support for Compose and dead-keys with libxkbcommon

### LINKED ISSUE #4: i3lock don't handle compose and dead-keys
Because of this reason, i3lock can't handle passwords with accents (e.g. á, ë) or other complex glyphs which are composed with 2 or more key presses.

This make the application pretty unusable for that kind of passwords.

danielotero:
This PR should resolve #4

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -334,7 +334,7 @@ static void handle_key_press(xcb_key_press_event_t *event) {
             /* xkb_compose_state_get_utf8 doesn't include the terminating byte in the return value
              * as xkb_keysym_to_utf8 does. Adding one makes the variable n consistent. */
             n = xkb_compose_state_get_utf8(xkb_compose_state, buffer, sizeof(buffer)) + 1;
-            //ksym = xkb_compose_state_get_one_sym(xkb_compose_state); // Not sure if it's needed
+            ksym = xkb_compose_state_get_one_sym(xkb_compose_state);
             composed = true;
             break;
         case XKB_COMPOSE_CANCELLED:
--- a/i3lock.c
+++ b/i3lock.c
@@ -339,7 +339,7 @@ static void handle_key_press(xcb_key_press_event_t *event) {
             break;
         case XKB_COMPOSE_CANCELLED:
             xkb_compose_state_reset(xkb_compose_state);
-            break;
+            return;
         }
     }
 
--- a/i3lock.c
+++ b/i3lock.c
@@ -25,6 +25,7 @@
 #include <ev.h>
 #include <sys/mman.h>
 #include <xkbcommon/xkbcommon.h>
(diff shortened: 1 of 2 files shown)


i3lock.c
```diff
@@ -289,13 +323,36 @@ static void handle_key_press(xcb_key_press_event_t *event) {
...
+        switch (xkb_compose_state_get_status(xkb_compose_state)) {
+        case XKB_COMPOSE_NOTHING:
+            break;
+        case XKB_COMPOSE_COMPOSING:
+            return;
+        case XKB_COMPOSE_COMPOSED:
+            /* xkb_compose_state_get_utf8 doesn't include the terminating byte in the return value
+             * as xkb_keysym_to_utf8 does. Adding one makes the variable n consistent. */
+            n = xkb_compose_state_get_utf8(xkb_compose_state, buffer, sizeof(buffer)) + 1;
+            //ksym = xkb_compose_state_get_one_sym(xkb_compose_state); // Not sure if it's needed
```
stapelberg: Without this line, people cannot compose return/enter/xf86screensaver, right? I think it should be in the code.

i3lock.c
```diff
@@ -138,6 +141,37 @@ static bool load_keymap(void) {
 }
 
 /*
+ * Loads the XKB compose table from the given locale.
+ *
+ */
+static bool load_compose_table(const char *locale) {
+    if (xkb_context == NULL) {
```
stapelberg: You can delete line 148 to 153, xkb_context is guaranteed to be non-NULL at this point (it’s initialized in `load_keymap` and i3lock exits if that fails).

stapelberg:
Very nice!

i3lock.c
danielotero: Ok. I just use `load_keymap` as a template, because I never use the libxkbcommon.

i3lock.c
danielotero: The function `xkb_keysym_to_utf8` is moved below and is executed only if the current ksym doesn't trigger the compose state machine. In fact, is executed even if the compose combination is cancelled, which is different from what X does. I should fix that too.

stapelberg: > The return/enter/xf86screensaver works for me as expected and I'm not sure that makes sense compose that kind of keys.

Why not? You never know what weird keyboard layouts people have… :)

> And I reallize that the key still gets readed after a failed composing attempt, which is different from what X does. I should fix that too.

I interpret that as you’re not quite done with this pull request yet, then.

danielotero: > Why not? You never know what weird keyboard layouts people have… :)

You are right. I never thought about a composing other than accents.

Now the line is uncommented and the behaviour is like the X server. If no other problem arises, I think it's done.

PR #7, opened 2015-03-03
Feature: clock

lexszero:
This pull-request adds a neat clock on locked screen. You can turn it on with -T option, and select font color with -F option.
This feature can be useful, for example, when you just want to know what time it is waking in the middle of the night and have just a laptop nearby.
How it looks for me with -c 000000 -F 00ff00 and active unlock indicator:
Screenshot

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -23,12 +23,15 @@ i3lock \- improved screen locker
 .IR image.png \|]
 .RB [\|\-c
 .IR color \|]
+.RB [\|\-F
+.IR fgcolor \|]
 .RB [\|\-t\|]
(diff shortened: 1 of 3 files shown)


stapelberg:
Sorry, but I don’t think we want to merge this. There are plenty of things that make sense to some people to display on the lockscreen, but we want to try hard to avoid feature creep in i3lock. If you want a lock screen that provides more features, use these provided by desktop environments like gnome.

stibi:
Hi @lexszero , I like it, so I'll build i3lock for me including this feature, thanks! Just so you know, that your work is not forgotten :+1: Thanks!

PR #14, opened 2015-04-26
Make pkg-config configurable

ony:
To fulfill needs of some linux distros like Exherbo which use target prefixed tools (like `x86_64-pc-linux-gnu-pkg-config`) please consider merging in this pull request.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,19 @@
 INSTALL=install
 PREFIX=/usr
 SYSCONFDIR=/etc
+PKG_CONFIG=pkg-config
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,18 +1,19 @@
 INSTALL=install
 PREFIX=/usr
 SYSCONFDIR=/etc
+PKG_CONFIG ?= pkg-config
```
stapelberg: Why `?=` instead of `=`? Can you please keep the line consistent with the lines above?

ony: With `?=` if you have `PKG_CONFIG` environment variable its value will be used instead of this one (`pkg-config`).

PR #25, opened 2015-07-18
Update unlock_indicator.c

vistamon82:
Changed line 204 
text = "verifying…";

to

text = "verifying...";

Unicode ellipsis shows up as a blank box, so changed to 3 periods.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,10 @@
+BasedOnStyle: google
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortBlocksOnASingleLine: false
+AlwaysBreakBeforeMultilineStrings: false
+IndentWidth: 4
+PointerBindsToType: false
+ColumnLimit: 0
+SpaceBeforeParens: ControlStatements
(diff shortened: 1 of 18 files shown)


Airblader:
As someone who actually cares about proper typography, seeing this PR hurts me, but ultimately I guess I have to agree. I'll just look the other way. :)

stapelberg:
> Unicode ellipsis sometimes shows up as a blank box, so changed to 3 periods.

Can you be more specific about what “sometimes” means? Could it be that we need to be more picky about which font we request from the system? Like @Airblader, I’m inclined to fix this the proper way :).

threefanger:
@Airblader @stapelberg I found that you had to have ttf-ms-fonts installed for the blank box to not show up.

@stapelberg i meant all the time, without the proper font.

stapelberg:
From http://sources.debian.net/src/cairo/1.14.0-2.1/src/cairo.c/?hl=2720#L2708:

```
 * If text is drawn without a call to cairo_select_font_face(), (nor
 * cairo_set_font_face() nor cairo_set_scaled_font()), the default
 * family is platform-specific, but is essentially "sans-serif".
```

I think there are two ways to go about this issue:

1. Find out why your distribution ships with a default font that does not contain the unicode glyph “…” and convince your distribution to ship a better one.
2. Make i3lock specifically request a widely-available font which contains the unicode glyph “…”.

PR #27, opened 2015-07-21
Features

0xbad0c0d3:
+control socket
+scale image if cairo version >=1.14.0
--display, --user features
--cmd - send command to i3lock-server using control socket
udev.rules
systemd.service


0xbad0c0d3:
So? What do you think?

Airblader:
I think that you have not at all reacted to the last feedback. It's still one PR consisting of multiple features and there is still hardcoding that breaks things (the version, for example). 

As for Michael's feedback you just need to wait as he is currently travelling.

stapelberg:
As explaind by @Airblader, we won’t merge a huge pull request which adds tons of features at once. If you’re serious about getting the features upstream, please file separate pull requests.

PR #37, opened 2015-10-07
Added background image scaling

brueggemann:
Implemented a background image scaling feature with new optional argument -s, because it was a little bit annoying, when changing to another screen with higher resolution at work.


stapelberg:
Sorry, i3lock does not accept patches for image manipulation of any kind. External programs should be used for that.

PR #39, opened 2015-10-08
Install manpage as default

cicku:


Airblader:
I expect this PR to be closed. Installing man pages is the job of the package maintainer as paths differ on different environments.

stapelberg:
What @Airblader said.

cicku:
Ok

Issue #12, opened 2015-04-23
make redrawing the screen optional when no unlock visual is used


phavx:
Former discussion here: https://faq.i3wm.org/question/5839/i3lock-how-to-disable-redraw/

start of the idea/problem here: http://redd.it/328e18

If you read this threads, you'll see I'm drawing some status information onto i3locks window with conky. The unlock indicator is disabled, so in theory there should be no need for a screen redraw (at least I see no technical need for this). As a result of redrawing the screen, my conky vanishes and reappers on every keypress (which looks strange and ugly) - happens with conky no matter if I used double buffer or not.

Therefor, I'd like to request an option to disable the redraw. Thanks.

stapelberg:
The `unlock_indicator` option that we already have should be used for this. Please send a pull request in which the `redraw_screen()` calls on key presses are skipped if `unlock_indicator` is disabled. It’s better if you do it because then you know whether it works.

Airblader:
@phavx Can you please try this patch?

http://pastebin.com/0vLwEgpU

phavx:
@Airblader tried your patch against a fresh clone of i3lock, applies without warnings and works for me. Thank you :)

Still getting this, don't know if it's because I'm not running a compositor or something else, but the patch works for me:

X Error of failed request:  BadDamage (invalid Damage parameter)
  Major opcode of failed request:  142 (DAMAGE)
  Minor opcode of failed request:  3 (XDamageSubtract)
  Serial number of failed request:  67
  Current serial number in output stream:  68

Airblader:
@stapelberg I'm getting the same error. I don't really know anything about the Damage extension. Is this normal?

I'll send a PR for the patch. Since i3lock releases are rare, I'd suggest you just go with the patch.

PR #41, opened 2015-10-17
add options to set margins for fullscreen window

bennofs:
This change adds four new options to set the top, left, right
and bottom margins of the fullscreen window. That is useful if
you want to have additional programs (such as docks) be visible
while the screen is locked.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -735,6 +735,10 @@ int main(int argc, char *argv[]) {
     struct passwd *pw;
     char *username;
     char *image_path = NULL;
+    int top_margin = 0;
+    int bottom_margin = 0;
+    int left_margin = 0;
+    int right_margin = 0;
     int ret;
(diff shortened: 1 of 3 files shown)


stapelberg:
This option provides a gun with which people can shoot themselves in the foot: by not covering the entire screen, sensitive information might be leaked, possibly in combination with your dock (or whatever other window you want to display) crashing or changing behavior in some other way.

It seems to me like issue #12 describes a more robust way of making information visible while your screen is locked: painting on the i3lock window.

I suggest you either use the mechanism outlined in issue #12 or publish this patch (or a fork) elsewhere — there are already a bunch of forks of i3lock with individual features that we think should not live in the official i3lock.

bennofs:
Ok, thank you for the detailed response anyway. I'll just patch i3lock locally then.

PR #42, opened 2015-10-18
Empty the event queue after pam fails to authenticate

seirl:
This avoids bad people repeating the "key + enter" combination, which puts
i3lock in a loop of pam validation delays and prevents the owner of the session
to log in properly before the whole queue is emptied.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -303,6 +303,16 @@ static void input_done(void) {
     if (unlock_indicator)
         redraw_screen();
 
+    /* Skip all the events during the pam verification to avoid bad people
+     * spamming keys and locking pam in an endless validation loop */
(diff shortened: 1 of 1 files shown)


stapelberg:
Why does commit 8a630bec651be1670f45c022a35399529ab0e523 not cover this problem?

I’m hesitant to merge your pull request as-is, because it breaks the use-case of noticing you had a typo in your password and re-entering it while PAM blocks.

seirl:
Hum, you're right, I wasn't testing on master.  8a630be already fixes this issue.

Airblader:
Thanks for letting us know. I'm going to close this PR then now for the reasons Michael mentioned. Thanks for your contribution, though!

PR #47, opened 2015-12-05
Display time at the center of the unlock circle

Ark444:
How frustrating to have to unlock the screen only to look at the time! I thus added this simple option (-C or --clock) to only have to press a key to get the time.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -30,6 +30,7 @@ i3lock \- improved screen locker
 .RB [\|\-I
 .IR timeout\|]
 .RB [\|\-f\|]
+.RB [\|\-C\|]
 
 .SH DESCRIPTION
 .B i3lock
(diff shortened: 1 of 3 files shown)


Airblader:
Thank you for the contribution. However, we generally advise people to open an issue for a feature first in order to have it discussed regarding whether this would be merged at all and if so, what it should look like etc. This way, we reduce the chance of anyone putting wasted effort into something.

Unfortunately, your PR fits into that scheme. We want to keep i3lock small to also keep it safe. Moreover, this opens the gate for feature creeping – you want the time, others the date, then locale information come into play and other people want to display email counts and so on. See #7 for a previous PR doing something similar and the reasoning.

For those reasons, I'd assume that @stapelberg is going to close this PR as in #7.

Ark444:
Alright, I understand, thank you for replying that fast. :)

Issue #36, opened 2015-09-21
[enhancement] Show locking state


Airblader:
See https://github.com/i3/i3lock/issues/35. i3lock should display "locking..." while it is trying to grab the pointer so the user knows when the locking was successful.

stapelberg:
Note that we should only start displaying that text after a little delay, maybe 250ms, so that in regular operations, people won’t see it (only if the operation is about to fail).

stapelberg:
Also note how we need to pay attention to forking: we should only fork when we could grab the pointer/keyboard, and the exit code still needs to be 0/1 according to success/failure of grabbing.

PR #48, opened 2015-12-06
Add support for grace period.

jmgao:
This PR adds support for a grace period on lock, for use with xautolock and the like. time_t overflow is irritating to handle in a portable fashion, so it's left unimplemented. This doesn't actually matter too much, since it'll only affect people who get autolocked before y2k38 and try to unlock afterwards. Likewise, anyone who wants to wait 136 years can unlock your screen if they time it properly.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -72,6 +72,11 @@ pressing Escape. Only makes sense together with \-d. If omitted, the default is
 30 seconds.
 
 .TP
+.BI \-g\  seconds \fR,\ \fB\-\-grace-period= seconds
+Specifies the number of seconds after startup i3lock will allow unlock without
+password. If omitted, or 0 is provided, i3lock will always require a password.
+
+.TP
 .B \-u, \-\-no-unlock-indicator
 Disable the unlock indicator. i3lock will by default show an unlock indicator
 after pressing keys. This will give feedback for every keypress and it will
--- a/i3lock.c
+++ b/i3lock.c
@@ -6,6 +6,7 @@
  * See LICENSE for licensing information
  *
  */
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <pwd.h>
@@ -46,6 +47,8 @@ typedef void (*ev_callback_t)(EV_P_ ev_timer *w, int revents);
 
 char color[7] = "ffffff";
 int inactivity_timeout = 30;
+time_t grace_period = 0;
+struct timespec start_time;
 uint32_t last_resolution[2];
 xcb_window_t win;
 static xcb_cursor_t cursor;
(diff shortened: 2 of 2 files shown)


stapelberg:
I’m a bit undecided about whether this feature should go into i3lock at all. Here are a couple of thoughts:

* i3lock strives to be minimal, and the mental model of “i3lock is started, my screen is locked” is very appealing. In general, we try hard to not even offer ways in which users can accidentally shoot themselves in the foot, and an option with which your screen is not actually locked could be considered such a feature.

* If you frequently accidentally lock your screen, it seems to me like the cleaner solution is to increase the interval with which xautolock triggers a lock, or to just always manually lock your screen. The grace period sounds like fighting the symptoms.

* If we decided to merge the feature, I think it should be implemented once issue #36 is resolved, i.e. once i3lock can tell the user what the current state is. The grace time would then be artificially prolonging the “locking” state.

jmgao:
> I’m a bit undecided about whether this feature should go into i3lock at all. Here are a couple of thoughts:

Is there any better place to put it? Putting it in xautolock is the only other reasonable place I can think of, but it never actually draws anything.

> i3lock strives to be minimal, and the mental model of “i3lock is started, my screen is locked” is very appealing. In general, we try hard to not even offer ways in which users can accidentally shoot themselves in the foot, and an option with which your screen is not actually locked could be considered such a feature.

You have to go out of your way to enable this option, and it's fairly self explanatory (at least to me), so it seems unlikely to me that someone would accidentally do this. Perhaps limiting it to "reasonable" values would be a good idea.

> If you frequently accidentally lock your screen, it seems to me like the cleaner solution is to increase the interval with which xautolock triggers a lock, or to just always manually lock your screen. The grace period sounds like fighting the symptoms.

Well, it is fighting the symptoms, but there are two diametrically opposed goals, so I'm not sure there's anything better. I want to minimize the amount of time my screen is unlocked, but I also don't want my screen to lock spuriously (or at least I want to be able to do something about it). A 60 second xautolock timeout is effectively the same as a 55 second timeout with a 5 second grace period, but the latter lets the user actually do something about it.

> If we decided to merge the feature, I think it should be implemented once issue #36 is resolved, i.e. once i3lock can tell the user what the current state is. The grace time would then be artificially prolonging the “locking” state.

This makes sense. If I have some spare time this week, I'll look into implementing that.

Airblader:
I think the usecase for this is pretty common. However, I could also see this done by a script wrapping i3lock. There was [this post](https://www.reddit.com/r/unixporn/comments/3v49i8/oc_i_made_a_thing/) that someone made a couple of days ago and the script could basically be reused to display a "Locking in 3…2…1…" message and only then calling i3lock.

Just as an alternative. I'm not entirely against putting this into i3lock itself, but I share the same concerns.

bdemirkir:
I agree with @jmgao and @Airblader.

It's very common use case. Right now I have to give 5 minues timeout before triggering i3lock just because I don't have any grace period option. I don't want to enter my password when reading a long article. With a grace period, I can safely lower it to 1 minute which is a pretty good security improvement from my point of view.

Maybe we can make use of xautolock's killer and killtime options. But it may get a little bit hacky. I rather prefer the same visual with i3lock.

pda:
> Likewise, anyone who wants to wait 136 years can unlock your screen if they time it properly.

Could this make life easier for an attacker that has control over the external NTP server being used by the locked system?

Airblader:
Only if you don't wait around to verify that your system has been locked. And if you don't do that, you also have the risk of physical attacks, which seems more likely than the NTP scenario to me anyway. So I don't think we need to consider that case.

The user just needs to know that they should wait until they get visual indication that the system has been locked.

jmgao:
> Could this make life easier for an attacker that has control over the external NTP server being used by the locked system?

CLOCK_MONOTONIC isn't affected by NTP discontinuities. I guess in theory, you could affect the speed of the system's clock, since [adjtime(3)](http://man7.org/linux/man-pages/man3/adjtime.3.html) affects CLOCK_MONOTONIC, but it slews the clock extremely gradually, so I don't think it's a useful attack.

stapelberg:
The [xss-lock manpage](https://www.mankier.com/1/xss-lock#Examples) demonstrates how this feature can be implemented without touching i3lock:

> Dim the screen after three minutes of inactivity, lock the screen two minutes later using i3lock:
> xset 180 120
> xss-lock -n dim-screen.sh -- i3lock -n

With that in mind, I think we should close this pull request. @jmgao, agreed?

Airblader:
I'm going ahead with what @stapelberg said now.

PR #49, opened 2015-12-10
Added password indicator so you can see how much of your password you…

mikeecb:
… have entered.

-h or --no-password-indicator to not show the password indicator.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -79,6 +79,13 @@ show you the current PAM state (whether your password is currently being
 verified or whether it is wrong).
 
 .TP
+.B \-h, \-\-no-password-indicator
+Disable the password indicator. i3lock will by default show an password indicator
+after pressing keys. This will give feedback for every keypress by showing you
+how much of the password you have already typed using the ● character up to a
+max length of 15 characters.
+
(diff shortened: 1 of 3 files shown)


stapelberg:
The maximum length of 15 characters that you needed to set is one of the main reason why i3lock doesn’t show how much you have entered.

What’s the use-case for this?

mikeecb:
Well that number is arbitrary. I believe the max length is 512 characters for password input but most input fields cap input length for aesthetics. 

Use case is similar to other feedback but less intrusive. So if you disable the circle (-u) you know if the password input is empty, or not. The 15 cap could be user defined as well. So if you mistype a letter you don't have to delete everything and start again. Quite common on most OS log in screens.

stapelberg:
> Use case is similar to other feedback but less intrusive. So if you disable the circle (-u) you know if the password input is empty, or not.

Understood — it comes down to personal preference.

I’ll decline this pull request, as I think we’re covered fairly well with the current unlock indicator. We want to keep the code complexity and mental complexity very low with i3lock intentionally, as it’s a very security-sensitive program.

PR #54, opened 2016-01-11
Increase performance when having alpha channel in png image

kbeckmann:
I had pretty big performance issues with i3lock on my laptop. Turns out that it's caused by PNGs with an alpha channel. By changing the operator when rendering the PNG the alpha channel is ignored and we get high performance.

I have only tested this code on my machine but it works well. CAIRO_OPERATOR_OVER is the default operator.

### DIFF
--- a/unlock_indicator.c
+++ b/unlock_indicator.c
@@ -116,8 +116,10 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
 
     if (img) {
         if (!tile) {
+            cairo_set_operator(xcb_ctx, CAIRO_OPERATOR_SOURCE);
             cairo_set_source_surface(xcb_ctx, img, 0, 0);
             cairo_paint(xcb_ctx);
+            cairo_set_operator(xcb_ctx, CAIRO_OPERATOR_OVER);
         } else {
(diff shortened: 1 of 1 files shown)


Airblader:
Why is the alpha channel ignored now? Shouldn't using the source operator cause a transparent image to be drawn, assuming a compositor is running? Also, what about the case where tiling is used for the image?

Airblader:
To clarify, I'm not saying respecting transparency is what should happen (it definitely isn't), but I'd just expect this to happen now.

kbeckmann:
Ok, please ignore this PR, i only considered my setup. This patch is not production quality, just a pointer in a general direction.

But there's definitely a lot of performance to gain, it seems cairo has some serious performance issues with alpha blending. On my lenovo yoga 3, each 1920x1080 frame took somewhere around 200-500ms to render which is totally unacceptable. That's why I wrote the patch. Maybe should've made an issue and post the code there instead.

stapelberg:
Isn’t this a cairo issue? Can you reproduce the issue with a minimum working example, i.e. not i3lock? If yes, please file a bug with cairo. If not, pull requests that don’t change the behavior but increase performance are appreciated.

I’m closing this PR for now.

PR #61, opened 2016-02-22
added ability to execute commands in case of wrong password (-E)

loadaverage:
Hi there.
Option to execute commands in the case of wrong password. Like send email or take a photo from webcam or just run bash script with some logic. 
`i3lock -E "echo \"Someone typed wrong password\" | mail -s oops cve@example.org"`
UPD: fixed indentation

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -79,6 +79,7 @@ cairo_surface_t *img = NULL;
 bool tile = false;
 bool ignore_empty_password = false;
 bool skip_repeated_empty_password = false;
+char *ext_exec = NULL;
 
 /* isutf, u8_dec © 2005 Jeff Bezanson, public domain */
 #define isutf(c) (((c)&0xC0) != 0x80)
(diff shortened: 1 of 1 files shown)


stapelberg:
I’m hesitant to merge this. Running arbitrary commands on failed unlock attempts can easily lead to security issues. I realize that users would need to enable this feature and it’s “their own fault” if they use a script/program which reduces security. Still, I think we might not want to encourage users.

Note that `--show-failed-attempts` covers the use-case in your example.

loadaverage:
I'm understand this. So, maybe add option to Makefile that enables this for those who understands all?
Anyway, I've created patches for upstream and latest versions [here](https://github.com/loadaverage/i3lock-wrongpass/tree/wrong_pass_exec/patches)
Thanx.

stapelberg:
I’m not a fan of compile-time optional features either :). We don’t have any of these for i3lock and i3status, and only a single one for i3 itself (-DPANGO_SUPPORT) which I’d like to eliminate eventually.

Given that the patch is tiny and we don’t plan to do big changes to i3lock at all, maybe the best course of action would be if you maintained this feature in a separate repository, as it looks like you already do?

Airblader:
(Well we do have a flag for cairo support as well but that too we want to remove once we've had maybe one or two releases with it and know that it works fine)

loadaverage:
> Given that the patch is tiny and we don’t plan to do big changes to i3lock at all, maybe the best course of action would be if you maintained this feature in a separate repository, as it looks like you already do?

@stapelberg yes, I think is the good solution :)

stapelberg:
Great. Closing this pull request in favor of your repository then.

PR #62, opened 2016-03-02
Added a command-line option to resize the background image

ZetsubouClown:
This patch allows the user to specify an option to resize the brackground image to completely fill the screen. At its current state when used with multi-monitor setup this will stretch the image across all screens, which may not be appropriate.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -93,6 +93,11 @@ If an image is specified (via \-i) it will display the image tiled all over the 
 (if it is a multi-monitor setup, the image is visible on all screens).
 
 .TP
+.B \-s, \-\-stretch
+If an image is specified (via \-i) it will resize the image to fit the screen
+(if it is a multi-monitor setup, the image fills all screens as if they were one).
+
+.TP
 .BI \-p\  win|default \fR,\ \fB\-\-pointer= win|default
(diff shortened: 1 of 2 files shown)


Airblader:
We don't want image processing inside i3lock. There's plenty of tools the user can use to generate an image that will work correctly and then pass that to i3lock.

In general we advise that you first open an issue for a feature request to avoid working on something we won't merge.

ZetsubouClown:
What about having multiple screens with different resolutions? As I understand it, with this setup there is no way to show the image on both screens without tiling. Or is this just something that i3lock isn't supposed to do?

Airblader:
You just create an image with the size of the root window and use that. That'll work fine.

ZetsubouClown:
I know that, but my multi-headedness is not constant so that feature would be useful for me.
Alright, nevermind.

Airblader:
So create a script that creates that image dynamically. :) We just want to avoid image processing in i3lock to keep it simple (and therefore safer).

Airblader:
Just FYI, also see #37

ZetsubouClown:
Yeah, not searching for it was stupid of me, sorry.

PR #63, opened 2016-03-07
Add `login-failed-execute` argument

1u4nx:
My colleague is a bad boy and hacker, when I go to eating, he hacking my computer. 

So, I add a function, when password is wrong, execute a system command.

Such as:

i3lock --debug --login_failed_execute="mplayer -vo png -frames 1 tv://"

When the password is wrong, the camera automatically takes a photo.


stapelberg:
duplicate of #61

PR #64, opened 2016-03-21
Improve error message when an image cannot be used.

Steap:
Currently, if the image passed to i3lock through the "-i" option cannot be used
on the lock screen, then an error message is displayed, but it is not really
helpful. Display a more precise error message.

This commit adds a dependency on libmagic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ LIBS += $(shell $(PKG_CONFIG) --libs cairo xcb-dpms xcb-xinerama xcb-atom xcb-im
 LIBS += -lpam
 LIBS += -lev
 LIBS += -lm
+LIBS += -lmagic
 
 FILES:=$(wildcard *.c)
 FILES:=$(FILES:.c=.o)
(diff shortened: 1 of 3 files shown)


Airblader:
I'm not sure improving the error message is worth adding a new dependency.

Steap:
Indeed, but the current error messages are quite misleading, see for instance https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=776682 . The user would expect i3lock to tell them "nope, this is a jpg, you can only use a png", and instead they get "out of memory". Isn't there a way to be more explicit? I'm not sure whether cairo can do that.

stapelberg:
I think you should fix this in cairo — even if cairo can’t tell you what image type a file contains, it could tell you that it doesn’t contain a PNG that can be parsed by cairo. By fixing this in cairo instead of i3lock, you’re also helping a number of other applications.

Airblader:
@psychon Does that sound good to you?

Steap:
I sent a patch to cairo, as it indeed makes sense to fix the issue there:

https://lists.cairographics.org/archives/cairo/2016-March/027322.html

stapelberg:
Great, thank you so much! I’ll close this PR then.

psychon:
https://cgit.freedesktop.org/cairo/commit/?id=c5ee3f11b5fb5a652466bbe407d2dc8e0ef4eb71

PR #66, opened 2016-03-30
added option to scale images to screen size

biiiep:


Airblader:
Please see #62 or #37

biiiep:
ok, but why do you have this tiling feature. Isn't it the same risk?

biiiep:
then all this stuff should rather be handled by an external tool

Airblader:
> then all this stuff should rather be handled by an external tool.

Yes. imagemagick is perfectly capable of doing this.

> ok, but why do you have this tiling feature. Isn't it the same risk?

I don't know if there's another reason, but at the very least: compatibility. Sometimes we make mistakes and add features we later on would prefer to keep out, but once it's in and people use it, we don't want to break things for them.

biiiep:
> I don't know if there's another reason, but at the very least: compatibility. Sometimes we make mistakes and add features we later on would prefer to keep out, but once it's in and people use it, we don't want to break things for them.

Well if you see there a real security flaw you could really re-think removing the tiling feature and if not then even allowing adding new features in a "safe" way or at least a toolset for most of the usecases.

Or what do you think about adding some additional scripts/tools to handle this job. This would allow people to reuse ready and tested solution instead of building their own.

Airblader:
> Well if you see there a real security flaw you could really re-think removing the tiling feature

Like I said, we want to stay compatible. It's not so much that we see a security flaw right now, and maybe one PR like yours "would be fine". However, it gets tough arguing why one PR is accepted and another rejected (take this very discussion on the tiling feature as an example). And that leads to either frustration or feature creep.

i3lock is a screen locker and that's what it does. It's not a image manipulation program and it doesn't need to be, there are well-tested, widely available and known to work solutions for that out there that anyone can use exactly to their liking.

> Or what do you think about adding some additional scripts/tools to handle this job.

I don't think there's anything wrong with a `contrib/` folder and in i3 we have this. But there's also no reason you can't write such scripts or tools and distribute them yourself. Adding it into the main repository means that we suddenly have responsibility to maintain it, but we're not a very big project with a lot of manpower, so we try to avoid this. Just because someone contributes a script that works today doesn't mean they will keep maintaining it. But it's part of i3lock, so we're responsible for bug reports. That's not very satisfying and I'm not saying it's a showstopper, but ultimately, like I said, nothing is stopping you from writing scripts and distributing them…

… if they become popular, you will soon learn how everybody wants it to behave just a tiny bit different and you find yourself adding all kinds of options and custimizations and suddenly your simple script becomes a full-blown tool. :)

PR #72, opened 2016-04-28
Allow CTRL+J as enter and CTRL+H as backspace

### LINKED ISSUE #71: Allow Ctrl-J to validate the password
Hello guys,

First of all, thanks for this simple but perfect locker !

---

As a vim user, I like not to move my hands, for exemple to access the Enter key.. I usually use Ctrl-J to validate something, and would like to have that feature in i3lock also.

So for exemple, I type my password as usual, then I type Ctrl-J to validate the password

I went through the code, and think that there is something to change arround thoses lines : 
https://github.com/i3/i3lock/blob/master/i3lock.c#L369

What do you think about this ?

karulont:
fixes #71

Diff is a little messy because I had to increase indentation because of a additional if clause to check whether CTRL was also pressed.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -367,9 +367,13 @@ static void handle_key_press(xcb_key_press_event_t *event) {
     }
 
     switch (ksym) {
+        case XKB_KEY_j:
         case XKB_KEY_Return:
         case XKB_KEY_KP_Enter:
         case XKB_KEY_XF86ScreenSaver:
+            if (ksym == XKB_KEY_j && !ctrl)
+                break;
+
             if (pam_state == STATE_PAM_WRONG)
                 return;
 
@@ -414,7 +418,11 @@ static void handle_key_press(xcb_key_press_event_t *event) {
              * see issue #50. */
             return;
 
+        case XKB_KEY_h:
         case XKB_KEY_BackSpace:
+            if (ksym == XKB_KEY_h && !ctrl)
+                break;
+
             if (input_position == 0)
                 return;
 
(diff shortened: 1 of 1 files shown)


i3lock.c
```diff
@@ -367,22 +367,26 @@ static void handle_key_press(xcb_key_press_event_t *event) {
...
+        case XKB_KEY_j:
         case XKB_KEY_Return:
         case XKB_KEY_KP_Enter:
         case XKB_KEY_XF86ScreenSaver:
-            if (pam_state == STATE_PAM_WRONG)
-                return;
-
-            if (skip_without_validation()) {
-                clear_input();
+            if ((ksym == XKB_KEY_j && ctrl) ||
```
Airblader: I'd prefer if this was written as

````
if (ksym == XKB_KEY_j && !ctrl)
    break;

/* … */
```

That way we also avoid the indentation.

i3lock.c
```diff
@@ -414,21 +418,25 @@ static void handle_key_press(xcb_key_press_event_t *event) {
              * see issue #50. */
             return;
 
+        case XKB_KEY_h:
         case XKB_KEY_BackSpace:
-            if (input_position == 0)
-                return;
+            if ((ksym == XKB_KEY_h && ctrl) ||
```
Airblader: Same as above

Airblader:
Just to be careful, please (manually) verify that a password containing "j" can still be entered.

Airblader:
@stapelberg Should this behavior be documented somewhere?

karulont:
I am not real sure how to handle github pull requests. I don't know if there is a better way to update the pull request than to use git push -f. I also manually tested this with a password containing J, U, H keys.

Airblader:
@karulont Force pushing is the correct way. Github will update the pull request.

PR #74, opened 2016-05-15
Flush xcb connection after opening fullscreen window

### LINKED ISSUE #46: i3lock raise_loop process does not exit
In `raise_loop` i3lock is waiting for events on its window. On a few occasions this has caused i3lock to not exit on my system after unlocking. At best, this causes stray processes and at worst it prevents locking the screen when external scripts check whether the previous i3lock run terminated.

By analysing the running i3lock process using gdb I have found the window it is waiting for doesn't exist anymore (Obvisouly, this is expected as my session is unlocked). I'm going to run i3lock with `--debug` from now on so I get a better idea of what's going on.

I can't reproduce this yet.

sandsmark:
We need to ensure that the window handle is usable before returning so
it can be used immediately.

Fixes #46

### DIFF
--- a/xcb.c
+++ b/xcb.c
@@ -10,6 +10,7 @@
 #include <xcb/xcb.h>
 #include <xcb/xcb_image.h>
 #include <xcb/xcb_atom.h>
+#include <xcb/xcb_aux.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdbool.h>
@@ -149,6 +150,9 @@ xcb_window_t open_fullscreen_window(xcb_connection_t *conn, xcb_screen_t *scr, c
     /* Raise window (put it on top) */
     values[0] = XCB_STACK_MODE_ABOVE;
     xcb_configure_window(conn, win, XCB_CONFIG_WINDOW_STACK_MODE, values);
+
+    /* Ensure that the window is created and set up before returning */
+    xcb_aux_sync(conn);
 
     return win;
 }


xcb.c
```diff
@@ -150,6 +150,9 @@ xcb_window_t open_fullscreen_window(xcb_connection_t *conn, xcb_screen_t *scr, c
     values[0] = XCB_STACK_MODE_ABOVE;
     xcb_configure_window(conn, win, XCB_CONFIG_WINDOW_STACK_MODE, values);
 
+    /* Ensure that the window is created and set up before returning */
```
stapelberg: Strictly speaking, this will only ensure that the request to create the window is sent to X11, not necessarily that the window is created before returning. For the latter, you’d need `xcb_aux_sync(conn);`, and I believe this is actually what we should be using here, if I understand the issue correctly.

I think the `fork()` call marks a point where either the parent or the child process can be scheduled first. The child process opens a separate X11 connection and expects the window which the parent process created in the parent’s X11 connection using `open_fullscreen_window` to be present. Without the parent using `xcb_aux_sync(conn);`, this is not necessarily true because the X server could process the client’s new connection while the parent’s connection hasn’t even flushed out the buffered requests yet.

If I captured this correctly, could you also add that to the commit message please? I think such an intricate bug deserves a detailed description :).

sandsmark:
Thanks, what you wrote makes sense and from reading a bit about xcb_aux_sync() it seems like that is the proper solution.

stapelberg:
Can you also amend the commit description like I’ve suggested please?

stapelberg:
@sandsmark Ping? :)

sandsmark:
is the commit message ok now?

stapelberg:
Thanks!

PR #75, opened 2016-05-16
Add support for `xss-lock --transfer-sleep-lock'

cruegge:
`xss-lock` has an option to pass over the sleep lock fd from logind to the locker program, which is supposed to close it once the screen is locked. There's a contrib script in `xss-lock` to achieve that for `i3lock`, but it's probably cleaner to do this in the locker directly. This PR makes it possble to simply use

    xss-lock --transfer-sleep-lock -- i3lock --nofork

(`--nofork` is needed, since `xss-lock` expects that.)

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -628,6 +628,17 @@ static void xcb_prepare_cb(EV_P_ ev_prepare *w, int revents) {
 }
 
 /*
+ * Try closing logind sleep lock fd passed over from xss-lock, in case we're
+ * being run from there.
+ *
+ */
+static void maybe_close_xss_fd() {
+    char *xss_fd = getenv("XSS_SLEEP_LOCK_FD");
+    if (xss_fd)
+        close(atoi(xss_fd));
+}
+
+/*
  * Instead of polling the X connection socket we leave this to
  * xcb_poll_for_event() which knows better than we can ever know.
  *
@@ -671,6 +682,8 @@ static void xcb_check_cb(EV_P_ ev_check *w, int revents) {
                         exit(0);
 
                     ev_loop_fork(EV_DEFAULT);
+                } else {
+                    maybe_close_xss_fd();
                 }
                 break;
 
@@ -956,6 +969,7 @@ int main(int argc, char *argv[]) {
     if (pid == 0) {
         /* Child */
         close(xcb_get_file_descriptor(conn));
+        maybe_close_xss_fd();
         raise_loop(win);
         exit(EXIT_SUCCESS);
     }
--- a/i3lock.c
+++ b/i3lock.c
@@ -632,10 +632,15 @@ static void xcb_prepare_cb(EV_P_ ev_prepare *w, int revents) {
  * being run from there.
  *
  */
-static void maybe_close_xss_fd() {
-    char *xss_fd = getenv("XSS_SLEEP_LOCK_FD");
-    if (xss_fd)
-        close(atoi(xss_fd));
+static void maybe_close_sleep_lock_fd(void) {
+    const char *sleep_lock_fd = getenv("XSS_SLEEP_LOCK_FD");
+    char *endptr;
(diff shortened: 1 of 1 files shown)


i3lock.c
```diff
@@ -671,6 +682,8 @@ static void xcb_check_cb(EV_P_ ev_check *w, int revents) {
                         exit(0);
 
                     ev_loop_fork(EV_DEFAULT);
+                } else {
+                    maybe_close_xss_fd();
```
stapelberg: nit: Is it necessary to only close the file descriptor in the non-forking case, or could we just unconditionally close the file descriptor?

i3lock.c
```diff
@@ -628,6 +628,17 @@ static void xcb_prepare_cb(EV_P_ ev_prepare *w, int revents) {
 }
 
 /*
+ * Try closing logind sleep lock fd passed over from xss-lock, in case we're
+ * being run from there.
+ *
+ */
+static void maybe_close_xss_fd() {
```
stapelberg: nit: use `static void maybe_close_xss_fd(void) {` for functions which don’t expect parameters.

i3lock.c
```diff
@@ -628,6 +628,17 @@ static void xcb_prepare_cb(EV_P_ ev_prepare *w, int revents) {
...
+    char *xss_fd = getenv("XSS_SLEEP_LOCK_FD");
+    if (xss_fd)
+        close(atoi(xss_fd));
```
stapelberg: Can you use `strtol` with proper error checking instead of `atoi` please?

stapelberg:
While I’ve commented on technical details in the PR, let’s first evaluate whether doing this makes sense at all:

> There's a contrib script in xss-lock to achieve that for i3lock, but it's probably cleaner to do this in the locker directly. 

Can you elaborate on why exactly it’s cleaner to support xss-lock in i3lock directly? It seems to me that xss-lock could just support the forking process model which i3lock uses by default and then we would need neither a wrapper nor direct support in screen lockers.

I’m a bit hesitant to add code which only works in certain setups to avoid cluttering up our code with all sorts of special cases. This is especially important in security-related programs such as i3lock.

cruegge:
Not wanting unneccessary special code in i3lock is of course  a perfectly good argument against the PR.

I think xss-lock wants to support e.g. `loginctl unlock-session`, so they do not support the forking model. Moreover, I do not know which part of their logic depends on knowing whether the locker is alive or not. Maybe I'll go ask them.

The wrapper script essentially lets i3lock fork, closes the fd, and then does a `pkill -xu $EUID -0 i3lock` repeatedly to stay around until i3lock finished. On exit, it kills the locker using a `trap EXIT`. It's probably more of a subjective thing, but I do not fully trust shell scripts to not behave weirdly, and felt a bit better after getting rid of it -- even though this particular one is probably fine. In fact, I was basically happy to have figured that out, and thought I'd share in case you're interested.

I'll fix the nits depending on your opinion :)

stapelberg:
> think xss-lock wants to support e.g. loginctl unlock-session, so they do not support the forking model. Moreover, I do not know which part of their logic depends on knowing whether the locker is alive or not. Maybe I'll go ask them.

Ah, in order to kill i3lock upon unlock-session, xss-lock needs to know the pid, which is hard when i3lock forks — makes sense.

> The wrapper script essentially lets i3lock fork, closes the fd, and then does a pkill -xu $EUID -0 i3lock repeatedly to stay around until i3lock finished. On exit, it kills the locker using a trap EXIT. It's probably more of a subjective thing, but I do not fully trust shell scripts to not behave weirdly, and felt a bit better after getting rid of it -- even though this particular one is probably fine. In fact, I was basically happy to have figured that out, and thought I'd share in case you're interested.

I agree that handling this within i3lock is cleaner. Let’s go ahead with this PR.

cruegge:
It took a while (busy with other stuf), but I finally fixed your remarks.

PR #76, opened 2016-05-19
Read PNG from stdin when no argument is given

dopefishh:
This removes the need for intermediate files or hacky things like `/dev/stdin`. By changing `-i` to be optional and reading from stdin when no argument is given.

For example the following usecase, a blurred screenshot as lock image:
```
import -window root -resize 50% -resize 200% png:- | i3lock -i
```

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -743,6 +743,13 @@ static void raise_loop(xcb_window_t window) {
     }
 }
 
+cairo_status_t img_read(void *closure, unsigned char *data,
+                        unsigned int length) {
+    return fread(data, 1, length, stdin) == length ? CAIRO_STATUS_SUCCESS
+                                                   : CAIRO_STATUS_READ_ERROR;
+    (void)closure;
+}
+
 int main(int argc, char *argv[]) {
     struct passwd *pw;
     char *username;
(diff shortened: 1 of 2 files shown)


i3lock.c
```diff
@@ -811,7 +818,10 @@ int main(int argc, char *argv[]) {
                 unlock_indicator = false;
                 break;
             case 'i':
-                image_path = strdup(optarg);
+                if (optarg == NULL)
+                    image_path = "";
```
Airblader: The only reason this doesn't cause a crash is because we never free the string. Still not pretty, though.

dopefishh: Good point. Then I wonder, add flag and keep it `NULL` or `strdup("")`

Airblader: Let's wait to see whether @stapelberg would consider merging this in the first place. I just wanted to point it out already.

stapelberg:
I don’t understand why `/dev/stdin` is not considered good enough for this purpose? Could you elaborate please?

dopefishh:
- It's not pretty to use a virtual file for this purpose when it can also be done directly
- `/dev/stdin` is not POSIX, while this is not a very big issue there are systems that don't have it or only have it very recently(cygwin, some UNIX'es)
- reopening a file descriptor is just superfluous

Yes all these reasons are a bit artificial. [This](http://unix.stackexchange.com/questions/36403/portability-of-dev-stdout) and [this](http://unix.stackexchange.com/questions/123602/portability-of-file-descriptor-links) are pretty clear that it is very very very likely that all systems have it. However I still prefer a direct solution over a detour.

stapelberg:
The way I read your reply, you’re saying that subjectively, you like this implementation more than using `/dev/stdin`, but I don’t see any technical reasons why we should directly read from stdin, nor are there any previously impossible use-cases this enables.

Given that i3lock is a security-related program, I want to keep the code as small as possible (within reason). This specific addition does not seem worth the (admittedly small) complexity increase.

PR #83, opened 2016-07-09
Reread image on USR1 signal

mskorzhinskiy:
Many people use i3lock with images like so: make screenshot, apply some changes
changes to the image (like blur it or pixelize) and then, finally, call i3lock.

This applying sometimes takes a lot of time, especially with a lot of monitors
with a lot of pixels. So I've wrote this patch and use it as such:

1. Make a screenshot, lock the screen with it;
2. Blur the image a little bit;
3. Send USR1.
4. Goto (2).

And this allows me to make a beautifully looking locking process -- with blurring
screen step-by-step.

NOTE: This is my first patch to anything connected with ev and XCB, so if you like it
and will want to apply it review it carefully please.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -75,6 +75,9 @@ static struct xkb_compose_state *xkb_compose_state;
 static uint8_t xkb_base_event;
 static uint8_t xkb_base_error;
 
+struct ev_timer *img_reread_timer = NULL;
+
+char *image_path = NULL;
 cairo_surface_t *img = NULL;
 bool tile = false;
 bool ignore_empty_password = false;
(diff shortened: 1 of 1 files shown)


Airblader:
We don't really care about fancy pretitness with i3lock. First and foremost, it is a security applicaton and because of that, we keep it as simple as we can. Looks-related changes are generally denied for these reasons, so even though I use a fancy picture myself, I think with that argument we'd have to reject the PR.

However, I *could* imagine that the general functionality of reloading the picture on signal could be useful for other reasons. Hence I won't close this right away and rather hand the decision over to @stapelberg.

mskorzhinskiy:
OK, the reasons are quite clear.

But I can't imagine, beside fancy locking stuff, what this reasons are. Could you, please, describe it?

Airblader:
I don't have any reasons. I was just saying that I could imagine they exist.

stapelberg:
I can see the usefulness of such a modification for the use-case you describe. Further use-cases might include displaying the current time or other system information which might come in handy on a lockscreen.

In general, however, I think we should strive to have i3lock not change after being initialized. This covers both code and behavior: running as little initialization code as possible results in fewer places where things can go wrong — i.e. your screen being unlocked incorrectly! As for behavior, being able to leave your computer unattended and knowing that it will show the same picture when you return provides some piece of mind (to me at least…).

This pull request steers us away from the ideal I described above.

I think a better way to satisfy your use-case is to draw on i3lock’s window from an external process. That way, any modification is done entirely by the user, and any issues are confined to the user’s external process.

johnchen902:
> draw on i3lock’s window from an external process

Does anyone know which program do this?

luebking:
i3lock doesn't seem to grab the server, so you'll have to expose or find the lockscreen window id, open a GC on it and then draw random stuff there. As you would do as if the window was one of your own client - X11 allows every client to paint every drawable, charming and security hole =)

However, this approach has glitches because the next i3lock update will damage or erase your changes.

Not tried myself, but maybe it's possible to get an own override redirect window on top of i3lock (to eg. display a clock or similar)

luebking:
Update: dzen easily goes on top of i3lock, so if you want to add additional status, you can use this (or any other override_redirect window providing) client.

Makes me consider adding a clock and mpd status =)

PR #84, opened 2016-07-19
Removed obsolete inactivity timeout

zipperle:
The inactivity timeout is useless without DPMS

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -805,7 +805,7 @@ int main(int argc, char *argv[]) {
                 fprintf(stderr, "DPMS support has been removed from i3lock. Please see the manpage i3lock(1).\n");
                 break;
             case 'I': {
-                fprintf(stderr, "Inactivity timeout makes only sense with DPMS. Please see the manpage i3lock(1).\n");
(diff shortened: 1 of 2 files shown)


i3lock.c
```diff
@@ -806,10 +805,7 @@ int main(int argc, char *argv[]) {
                 fprintf(stderr, "DPMS support has been removed from i3lock. Please see the manpage i3lock(1).\n");
                 break;
             case 'I': {
-                int time = 0;
-                if (sscanf(optarg, "%d", &time) != 1 || time < 0)
-                    errx(EXIT_FAILURE, "invalid timeout, it must be a positive integer\n");
-                inactivity_timeout = time;
+                fprintf(stderr, "Inactivity timeout makes only sense with DPMS. Please see the manpage i3lock(1).\n");
```
stapelberg: Please add “, which was removed” so that people know what’s up without having to consult the manpage.

stapelberg:
Good catch!

i3lock.c
ixjlyons: Also a grammar nitpick: the word "only" should be moved. Either "...timeout *only* makes sense with..." or "...timeout makes sense *only* with..." would work, with a small personal preference for the former.

PR #85, opened 2016-07-27
add -s/--scaled option to scale the image on all available screens

billiob:
I was not satisfied with the image options and decided to add on to scale the input image on all screens. Without `-t`, the image is only displayed as-is.
This new option is incompatible with the `-t` option.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -84,6 +84,11 @@ If an image is specified (via \-i) it will display the image tiled all over the 
 (if it is a multi-monitor setup, the image is visible on all screens).
 
 .TP
+.B \-s, \-\-scaled
+If an image is specified (via \-i) it will display the image scaled on each
+available screen.
+
+.TP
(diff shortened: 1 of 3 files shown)


benibr:
Cool, I postpone to do this for a long time. Thanks!

Airblader:
#62, #66, #37

ivanhercaz:
Hi @Airblader! I am using i3lock and I really love it. I am configuring it to have an image but I looked that the image doesn't scale automatically, so searching for that option in the net I found this issue and read the explanations given in the three issues you referred above. I understand that you, the i3lock's maintainers, prefer to not include functions that might be done with external tools. However I think it would be useful to note in the README, may be in the second point of the introduction, that i3lock don't manipulate images nor plan to make it. I think it might be useful due it seems there are many people searching for this option in i3lock. If you are agree, I can propose a PR with that change.

Airblader:
Sure, sounds good, feel free to send the PR!

PR #88, opened 2016-08-16
Displaying locking message when grabbing the pointer.

### LINKED ISSUE #35: Right-click menu cancels lock
1. Open a right-click menu in some program.
2. Launch i3lock with relevant keybinding.
3. Lock screen appears for a few seconds before dropping user back to session.

Tested on i3lock 2.6.

### LINKED ISSUE #36: [enhancement] Show locking state

eplanet:
This PR fixes #36. Instead of waiting 250 ms to display the "locking…" message, I propose to display it right away since `START_TIMER` won't work when grabbing the pointer. However, I added a timer to hide the message after 250 ms in case of success. It actually makes sense to me to display the locking message each time.

Also, I shifted fork code down so the return value will actually be 1 in case of failure to grab the pointer.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -976,9 +976,15 @@ int main(int argc, char *argv[]) {
     /* Pixmap on which the image is rendered to (if any) */
     xcb_pixmap_t bg_pixmap = draw_image(last_resolution);
 
-    /* open the fullscreen window, already with the correct pixmap in place */
+    /* Open the fullscreen window, already with the correct pixmap in place */
     win = open_fullscreen_window(conn, screen, color, bg_pixmap);
     xcb_free_pixmap(conn, bg_pixmap);
+
+    cursor = create_cursor(conn, screen, win, curs_choice);
+
+    /* Display the "locking…" message while trying to grab the pointer/keyboard. */
+    pam_state = STATE_PAM_LOCK;
+    grab_pointer_and_keyboard(conn, screen, cursor);
 
     pid_t pid = fork();
     /* The pid == -1 case is intentionally ignored here:
@@ -992,9 +998,6 @@ int main(int argc, char *argv[]) {
         exit(EXIT_SUCCESS);
     }
 
-    cursor = create_cursor(conn, screen, win, curs_choice);
-
-    grab_pointer_and_keyboard(conn, screen, cursor);
     /* Load the keymap again to sync the current modifier state. Since we first
      * loaded the keymap, there might have been changes, but starting from now,
      * we should get all key presses/releases due to having grabbed the
@@ -1005,6 +1008,10 @@ int main(int argc, char *argv[]) {
     main_loop = EV_DEFAULT;
     if (main_loop == NULL)
         errx(EXIT_FAILURE, "Could not initialize libev. Bad LIBEV_FLAGS?\n");
+
+    /* Explicitly call the screen redraw in case "locking…" message was displayed */
+    pam_state = STATE_PAM_IDLE;
+    redraw_screen();
 
     struct ev_io *xcb_watcher = calloc(sizeof(struct ev_io), 1);
     struct ev_check *xcb_check = calloc(sizeof(struct ev_check), 1);
--- a/xcb.c
+++ b/xcb.c
@@ -18,8 +18,10 @@
 #include <unistd.h>
 #include <assert.h>
 #include <err.h>
+#include <time.h>
 
 #include "cursors.h"
+#include "unlock_indicator.h"
 
 xcb_connection_t *conn;
 xcb_screen_t *screen;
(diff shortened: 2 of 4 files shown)


stapelberg:
I would prefer it if the message was not displayed until necessary, to keep visual clutter at a minimum. Could you change the PR to do that please?

eplanet:
That might be possible, but I need some time to figure out how since the ev main loop must be started to handle the `ev_timer`, which is not yet started when the grabbing function is called. I wouldn't be against a clue if you've got some.

eplanet:
Here's a proposition, since we cannot use a `ev_timer` while grabbing the pointer, why bother ? We can display locking message, and hide it right away when the pointer has been grabbed. Testing this solution on my machine led to not displaying locking message at all when pointer was grabbed, and displayed locking message when failing to grab the pointer.

Is this solution acceptable for you or do you still want to dig further for displaying the locking message after 250ms of unfruitful grabbing attempt ?

stapelberg:
I think the current approach of displaying the locking message by default and hiding it once the pointer has been grabbed is indeed better than the previous approach, but there’s still room for improvement:

My concern is that on heavily loaded (or older) machines, or on setups where graphics drivers are facing some sort of issue (way more common than one would hope), the current approach will lead to flickering or the message being visible briefly. Personally, I’d like to trade some complexity for ensuring i3lock is visually appealing for all of our users.

So, yes, could you please try and find an approach which only shows the message when necessary? Thanks!

eplanet:
I think I found an acceptable solution. While grabbing the pointer, the code repeatedly checks elapsed time, and if it reaches 250 ms, it triggers a `redraw_screen()` call to display "locking…" message.

It preserves the return value before the fork if the grabbing fails, and while avoiding unnecessary "locking…" display is neither time consuming nor involves complicated code pattern.

If you accept this, I'll rebase to the master.

stapelberg:
The approach looks good to me! Thanks for refining this PR.

i3lock.c
```diff
@@ -976,10 +977,16 @@ int main(int argc, char *argv[]) {
...
     xcb_pixmap_t bg_pixmap = draw_image(last_resolution);
 
-    /* open the fullscreen window, already with the correct pixmap in place */
+    /* Open the fullscreen window, already with the correct pixmap in place */
     win = open_fullscreen_window(conn, screen, color, bg_pixmap);
     xcb_free_pixmap(conn, bg_pixmap);
 
+    cursor = create_cursor(conn, screen, win, curs_choice);
+
+    /* Setting to locking state while grabbing the pointer. */
```
stapelberg: The comment describes what we can already read from the code. Could you make it more informative by describing the effects? E.g.: Display the “locking” message while trying to grab the pointer/keyboard

i3lock.c
```diff
@@ -1006,6 +1010,9 @@ int main(int argc, char *argv[]) {
     if (main_loop == NULL)
         errx(EXIT_FAILURE, "Could not initialize libev. Bad LIBEV_FLAGS?\n");
 
+    pam_state = STATE_PAM_IDLE;
+    START_TIMER(clear_pam_lock_timeout, TSTAMP_N_SECS(0.25), redraw_timeout);
```
stapelberg: What is this timer used for?

xcb.c
```diff
@@ -170,6 +172,10 @@ void grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
 
     int tries = 10000;
 
+    /* Using few variables to trigger a redraw_screen() if too many tries */
+    bool redrawn = false;
+    time_t start = clock(), end;
```
stapelberg: This reads a bit awkward. Can you change it so that there is one variable definition per line please?

i3lock.c
eplanet: You're right, that timer is useless, it could be a simple call to `redraw_screen()`. That call should be needed since it can take some time before successfully grabbing the pointer and therefore we would need to clear the subsequent "locking..." message.

xcb.c
eplanet: I understand it can seem like some kind of abstruse syntactic sugar, I'll change that indeed.

eplanet:
Ok I've done the required corrections, thanks for the feedback. Feel free to ask further modifications if needed.

eplanet:
@stapelberg: what do you think?

eplanet:
@stapelberg : do you validate the modifications ?

stapelberg:
Thanks, I’ll take a closer look tomorrow.

xcb.c
```diff
@@ -190,6 +197,15 @@ void grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
 
         /* Make this quite a bit slower */
         usleep(50);
+
+        /* Measure elapsed time and trigger a screen redraw if elapsed > 250000 */
+        if (!redrawn && tries % 100 == 0) {
```
stapelberg: Please change the 3 lines into:

```c
if (!redrawn &&
    (tries % 100) == 0 &&
    (clock() - start) > 250000) {
```

eplanet: That totally makes sense. I modified the source file.

stapelberg:
Thanks for your contribution!

eplanet:
My pleasure, it was good fun to dig into i3lock sources! Thanks for your time and advice ;-)

PR #92, opened 2016-09-11
includes a flag -g for blurring a screenshot of your desktop, used al…

lvitals:
…one this will attempt to capture the entire screen contents and blur it. however, this may also be used in conjunction with both a loaded image file and the tiling option

### DIFF
--- a/blur.c
+++ b/blur.c
@@ -1,0 +1,128 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * © 2016 Leandro Vital
+ *
+ * See LICENSE for licensing information
+ *
+ */
+
+#include <math.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <cairo.h>
+
+#define ARRAY_LENGTH(a) (sizeof (a) / sizeof (a)[0])
+
+/* Performs a simple 2D Gaussian blur of radius @radius on surface @surface. */
+void
+blur_image_surface (cairo_surface_t *surface, int radius)
+{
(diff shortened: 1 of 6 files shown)


Airblader:
We don't want this kind of feature in i3lock because we want to keep it simple and easy to avoid security issues, so unfortunately I'll have to reject and close this PR. There's been many pull requests for the same and similar features where we explain this in more detail. We think this kind of processing is best placed outside of i3 with dedicated graphics software.

lvitals:
I understand this concern, like this kind of minimalist approach, but it is an improvisation using an external software for this, shell script is an improvisation not a solution. It is not a simple effect , UX Design is also necessary.

Airblader:
i3lock doesn't try to be fancy in any way. If you want a fancy-looking lock screen you may want to use something else. I also think that using something like imagemagick for this perfectly fits the UNIX philosophy.

lvitals:
i3lock has its fancy effects. It's a different screen lock of others. Unix philosophy is very good, small is beautiful, make each program do one thing well. But i3lock has other flags , eg "-i -c -b" native expanding its functionality , the blur effect is no different. I applied this functionality because I wanted my file ".i3/config" cleaner without many scripts.

Airblader:
We generally have some features that we regret having added. But we care about compatibility and cannot remove them now. That's also why we're careful with introducing new features. 

You can keep your config small by putting the command chain into a script. There's really nothing wrong with that and it's certainly no reason to bloat i3lock. You want this flag, I want others and somebody else yet another thing. That's feature creep.

PR #94, opened 2016-09-15
Add option to disable the state messages

johrpan:
... for those who don't like them.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -27,6 +27,7 @@ i3lock \- improved screen locker
 .IR pointer\|]
 .RB [\|\-u\|]
 .RB [\|\-e\|]
+.RB [\|\-s\|]
 .RB [\|\-f\|]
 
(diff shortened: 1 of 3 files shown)


Airblader:
I'll leave it to @stapelberg to decide about merging this at all, but if we do, I'd personally vote against `-s`. Not every option needs a single-letter alias and IMHO they should only be used for options that are intuitively clear. Worst case we end up with something much more fit for `-s` but then it's already used for this (very minor) change.

stapelberg:
Please just use `--no-unlock-indicator` instead. We don’t want to add hundreds of options to i3lock to make every little detail of it configurable.

Airblader:
Same time. :-)

johrpan:
Ok, no problem, luckily this is so small that I can just patch it for my own machine, but I thought there where others who would like to use this.

Issue #22, opened 2015-07-01
Bug: pop-up notifications in Chromium appear over i3lock screen


aliceriot:
When screen is locked using i3lock new notifications from Chromium appear on the screen above i3lock and are readable by any passers-by.

This happens whenever Chromium is running, the screen is locked, and Chromium displays a pop-up notification.

Possible workarounds are just disabling Chromium notifications, but in my mind i3lock should be blocking stuff like this. I haven't done extensive testing to see if spawning other types of windows or window-like things (while the screen is locked) will result in similar behavior.

Airblader:
Are you running a compositor?

aliceriot:
Yes, I use compton (b/c I like shiny things). Could this be the source of the problem?

Airblader:
Can you reproduce it without Compton?

aliceriot:
No, if I kill compton the behavior doesn't happen. Any ideas on how to work around this? I'd like to keep drop shadows and notifications without leaking information to anyone who sees my locked screen.

Airblader:
Sorry, but as stated in #8, we don't support any kind of compositing (though if you can come up with a PR, we certainly don't mind a fix for it), hence I'll close this issue.

I suggest you toy around with excluding any kind of compton effects on those popups and see if that changes something. Otherwise, try another lockscreen program. Also, if you do investigate this and can at least come up with the technical reason for why this happens, please don't hesitate sharing it with us – maybe a fix wouldn't be so hard.

Airblader:
@aliceriot By the way, what are these popups? Is it actually Chromium? Or is it some notification daemon? If you happen to be running dunst, it allows pausing / resuming the daemon which you can wrap around the i3lock call to fix this issue. Other daemons might have similar functionality.

aliceriot:
The popups are the chromium notification engine, gmail and such can send notifications which look much like desktop notifications using the framework. Technical document about them [here](https://developer.chrome.com/extensions/notifications). 

I'll definitely poke around at this! No promises on a PR though :D

Thanks for your help!

Airblader:
If you just find out what the underlying problem is, without a PR, please share that too. Maybe we can figure something out then. :9

Manouchehri:
I was having a similar issue with i3bar showing on top of i3lock (via a modifier key trigger) a few months ago. It's been fixed, not sure which commit though. 

Those two *might* be related, hopefully it'll give a starting place to look.

Airblader:
[...]

PR #96, opened 2016-09-25
Fix covering of composited notifications

sandsmark:
Use the XComposite extension to get the composite overlay window,
instead of just using the normal root window. This ensures that
composited windows are covered.

Fixes issue #22

### DIFF
--- a/xcb.c
+++ b/xcb.c
@@ -11,6 +11,7 @@
 #include <xcb/xcb_image.h>
 #include <xcb/xcb_atom.h>
 #include <xcb/xcb_aux.h>
+#include <xcb/composite.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdbool.h>
@@ -102,6 +103,29 @@ xcb_window_t open_fullscreen_window(xcb_connection_t *conn, xcb_screen_t *scr, c
     uint32_t mask = 0;
     uint32_t values[3];
     xcb_window_t win = xcb_generate_id(conn);
+    xcb_window_t parent_win = scr->root;
+
+    /* Check whether the composite extension is available */
+    const xcb_query_extension_reply_t *extension_query = NULL;
+    xcb_generic_error_t *error = NULL;
+    xcb_composite_get_overlay_window_cookie_t cookie;
+    xcb_composite_get_overlay_window_reply_t *composite_reply = NULL;
+
+    extension_query = xcb_get_extension_data(conn, &xcb_composite_id);
+    if (extension_query && extension_query->present) {
+        /* When composition is used, we need to use the composite overlay
+         * window instead of the normal root window to be able to cover
+         * composited windows */
+        cookie = xcb_composite_get_overlay_window(conn, scr->root);
+        composite_reply = xcb_composite_get_overlay_window_reply(conn, cookie, &error);
+
+        if (!error && composite_reply) {
+            parent_win = composite_reply->overlay_win;
+        }
+
+        free(composite_reply);
+        free(error);
+    }
 
     if (pixmap == XCB_NONE) {
         mask |= XCB_CW_BACK_PIXEL;
@@ -123,8 +147,8 @@ xcb_window_t open_fullscreen_window(xcb_connection_t *conn, xcb_screen_t *scr, c
 
     xcb_create_window(conn,
                       XCB_COPY_FROM_PARENT,
-                      win,       /* the window id */
-                      scr->root, /* parent == root */
+                      win, /* the window id */
+                      parent_win,
                       0, 0,
                       scr->width_in_pixels,
                       scr->height_in_pixels, /* dimensions */
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
     - libev-dev
     - libxcb-xinerama0-dev
     - libxcb-xkb-dev
+    - libxcb-composite0-dev
 before_install:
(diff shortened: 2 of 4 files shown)


sandsmark:
I guess it should be squashed before merging...

Airblader:
Yes, but first we should discuss whether it will be merged at all. The i3 suite generally doesn't support compositing and doesn't depend on compositing. On the other hand it's definitely a security related change. 

How does this behave if the X server is compiled without the Composite extension?

sandsmark:
the call should return an error which the code handles by just reverting to the old behavior, but I haven't tested it.

stapelberg:
Please let us know once you’ve tested it and I’ll have a look at this PR.

sandsmark:
I'll try to find time to do that.

sandsmark:
Tested with a xorg-server built with --disable-composite.

sandsmark:
I also changed the Makefile to make it build with libxcb built with --disable-composite.

stapelberg:
Thanks for testing.

> I also changed the Makefile to make it build with libxcb built with --disable-composite.

Is that a common situation or do major distributions ship with libxcb-composite? In general, we don’t use optional dependencies in i3*.

sandsmark:
I don't know of any distributions that doesn't ship with libxcb-composite. Even Slackware (the most conservative distro I could think of right now) ships it in their xcb package.

I did it optionally because of Airblader's comment about not depending on compositing.

stapelberg:
> I did it optionally because of Airblader's comment about not depending on compositing.

That comment meant (or at least I think it should have meant) depending on compositing being enabled at runtime.

Airblader:
Yes, what I had in mind was VNC as we had problems with extensions with those before.

sandsmark:
FWIW, xcb_composite_release_overlay_window() is not necessary, the overlay window is implicitly released when the X11 connection is terminated; https://www.x.org/archive/X11R7.5/doc/compositeproto/compositeproto.txt

stapelberg:
The changes in your pull request look good, but I’m wondering whether we’ve covered everything we need to cover. It seems like the basic change here is that when a compositor is used, the i3lock window no longer should be a child of the root window, but rather a child of the compositor window.

Does that mean we also need to query the compositor window’s geometry in `handle_screen_resize()`? (https://github.com/google/xsecurelock/commit/ee1825a4a113bd9a24bf8a499fe579d9a03d7612 seems to do that via the ConfigureNotify event).

Do we need to set the `XCB_EVENT_MASK_STRUCTURE_NOTIFY` event mask on the compositor window instead of the root window? (line 963)

Do we need to create background pixmaps on the compositor window instead of the root window? (`create_bg_pixmap()`)

And lastly, do we need grab pointer and keyboard on the compositor window instead of the root window? (`grab_pointer_and_keyboard()`)

karulont:
@stapelberg 's questions are valid.
I did some stuff with composite extension for [i3lock-blur](https://github.com/karulont/i3lock-blur). From what I remember the overlay window is special in some ways. I can't answers those questions anymore, but they are rights questions. I also know, that i3lock-blur has problems with composers like compton, thus I think that i3lock-blur can also benefit from those answers.

sandsmark:
To my understanding; no to all the questions. The composite overlay window is a very special "window". Please see section 3.2 in the document I referred to above.

I have also tested this patch with e. g. launching a compositor after launching i3lock and then launching clients that also try to grab various stuff.

As for xsecurelock, it does several weird things like e. g. releasing the composite window manually.

stapelberg:
Before merging this, I’ll need to read up on how precisely a compositor works (or can work, as per the spec), and what assumptions compositors might make. I’ll update this PR when I get around to that.

rickybrent:
If it gets merged in or not, the patch works great for me, thanks. :)

xcb.c
```diff
@@ -102,6 +103,29 @@ xcb_window_t open_fullscreen_window(xcb_connection_t *conn, xcb_screen_t *scr, c
     uint32_t mask = 0;
     uint32_t values[3];
     xcb_window_t win = xcb_generate_id(conn);
+    xcb_window_t parent_win = scr->root;
+
+    /* Check whether the composite extension is available */
+    const xcb_query_extension_reply_t *extension_query = 0;
+    xcb_generic_error_t *error = 0;
+    xcb_composite_get_overlay_window_cookie_t cookie;
+    xcb_composite_get_overlay_window_reply_t *composite_reply = 0;
```
Airblader: Can we please use `NULL` for the pointers? That way it's consistent with the rest of the codebase.

sandsmark: fixed

Airblader:
LGTM. Up to @stapelberg now :-)

stapelberg:
There isn’t much prior art here — aside from xsecurelock, no other screen locker seems to consider the composite extension.

I’ll go ahead and merge this code, but of course reserve the right to revert it later in case it causes issues or security vulnerabilities.

divVerent:
To add something to this issue - it seems like it's outright impossible for a screen locker to work with all compositors. For compton there are option combinations where this works (e.g. the default flags, or --backend glx --paint-on-overlay), for cinnamon's compositor it just works - but the spec correctly states that the result of two compositors interacting depends a lot on how the compositor works, and I'm considering making the feature either opt-in or only activate if there is _no_ compositor (or maybe a whitelisted one) in xsecurelock.

Issue #98, opened 2016-10-06
If you start drag but never drop with mouse, this will kill i3lock!!


artheus:
I consider this to be a security issue of the i3lock.

If someone would connect a secondary mouse to my computer at work. And whenever I lock my computer they will click the mouse button and not release it, this will kill i3lock after a little while.

So steps to reproduce this:

1. Click mouse button (start drag)
2. Start i3lock
3. w/o releasing the mouse button, wait for a little while
Boom! i3lock dies!

debug output is:
```
[i3lock-debug] device = 3
[i3lock-debug] found Xinerama screen: 1920 x 1080 at 0 x 0
[i3lock-debug] found Xinerama screen: 1920 x 1200 at 1920 x 0
[i3lock-debug] scaling_factor is 1, physical diameter is 190 px
[i3lock-debug] Watching window 0x01e00003
i3lock: Cannot grab pointer/keyboard
[i3lock-debug] Read event of type 18
[i3lock-debug] UnmapNotify for 0x01e00003
```

stapelberg:
This should be fixed with commit https://github.com/i3/i3lock/commit/1c97a8484723b7db6c96c3b7bfa4c86c26f47b5a already. Could you confirm please?

artheus:
I cloned the latest master, and build it locally.
```
$ i3lock -v
i3lock: version 2.8-7-g1c97a84 (2016-09-27) © 2010 Michael Stapelberg
```

This is the version I got from doing that.
I can not really understand if the `g1c97a84` part should be pointing to the latest commit. But from doing a git log I can see that the fix of issue #88 is the latest commit.

This do not seem to fix the problem though. Following the exact same steps from my first comment  in this issue results in the same issue, killing i3lock.

artheus:
My setup is Fedora
```
$ uname -a
Linux HOSTNAME 4.7.2-201.fc24.x86_64 #1 SMP Fri Aug 26 15:58:40 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```
(only replaced the hostname)

I use i3 as window manager (ofcourse)
```
$ i3 --version
i3 version 4.12 (2016-03-06, branch "4.12") © 2009 Michael Stapelberg and contributors
```

Airblader:
It's not supposed to stop i3lock from preventing. We simply cannot force locking because X doesn't allow us to lock, this i3lock gives up. The referenced commit is simply supposed to give you feedback on whether the screen is already locked or still in the process of attempting it.

[...]

PR #99, opened 2016-10-08
Display error message when locking failed

eplanet:
This PR is an attempt to address issue #98, displaying an error message before exiting when the pointer couldn't be grabbed and then the locking failed.

There's currently a problem with the text size that overflows the circle, I need to figure out how to split the text between two lines…

### DIFF
--- a/unlock_indicator.c
+++ b/unlock_indicator.c
@@ -160,6 +160,7 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
                 cairo_set_source_rgba(ctx, 0, 114.0 / 255, 255.0 / 255, 0.75);
                 break;
             case STATE_PAM_WRONG:
+            case STATE_I3LOCK_LOCK_FAILED:
                 cairo_set_source_rgba(ctx, 250.0 / 255, 0, 0, 0.75);
                 break;
             default:
@@ -174,6 +175,7 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
                 cairo_set_source_rgb(ctx, 51.0 / 255, 0, 250.0 / 255);
                 break;
             case STATE_PAM_WRONG:
+            case STATE_I3LOCK_LOCK_FAILED:
                 cairo_set_source_rgb(ctx, 125.0 / 255, 51.0 / 255, 0);
                 break;
             case STATE_PAM_IDLE:
@@ -212,6 +214,9 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
                 break;
             case STATE_PAM_WRONG:
                 text = "wrong!";
+                break;
+            case STATE_I3LOCK_LOCK_FAILED:
+                text = "lock failed!";
                 break;
             default:
                 if (show_failed_attempts && failed_attempts > 0) {
--- a/unlock_indicator.h
+++ b/unlock_indicator.h
@@ -11,10 +11,11 @@ typedef enum {
 } unlock_state_t;
 
 typedef enum {
-    STATE_PAM_IDLE = 0,   /* no PAM interaction at the moment */
-    STATE_PAM_VERIFY = 1, /* currently verifying the password via PAM */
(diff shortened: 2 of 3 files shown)


unlock_indicator.h
```diff
@@ -14,7 +14,8 @@ typedef enum {
     STATE_PAM_IDLE = 0,   /* no PAM interaction at the moment */
     STATE_PAM_VERIFY = 1, /* currently verifying the password via PAM */
     STATE_PAM_LOCK = 2,   /* currently locking the screen */
-    STATE_PAM_WRONG = 3   /* the password was wrong */
+    STATE_PAM_WRONG = 3,  /* the password was wrong */
+    STATE_PAM_FAIL = 4    /* display error message before exiting */
```
Airblader: The comment on this value should be what it represents, not what should be displayed for it.

unlock_indicator.c
```diff
@@ -213,6 +215,9 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
             case STATE_PAM_WRONG:
                 text = "wrong!";
                 break;
+            case STATE_PAM_FAIL:
+                text = "Failed; exiting";
```
Airblader: I think we should improve this message.

xcb.c
```diff
@@ -233,8 +235,12 @@ void grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
         }
     }
 
-    if (tries <= 0)
+    if (tries <= 0) {
+        pam_state = STATE_PAM_FAIL;
+        redraw_screen();
+        usleep(2000000);
```
Airblader: Given that it's not obvious I'd prefer a comment here explaining why there's a sleep before exiting.

unlock_indicator.c
eplanet: Indeed. I'd rather go for a simple "Locking failed", but still is out of bounds. Do you think we could go for a display on two lines ?

eplanet:
I corrected some details according to @Airblader feedback. I think the PR can be merged from now, if you're happy with it.

xcb.c
```diff
@@ -233,8 +235,14 @@ void grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
         }
     }
 
-    if (tries <= 0)
+    /* After trying for 10000 times, i3lock will display an error message
+     * for 2 sec prior to terminate. */
+    if (tries <= 0) {
+        pam_state = STATE_PAM_FAIL;
+        redraw_screen();
+        usleep(2000000);
```
stapelberg: Use sleep() instead of usleep() when you’re specifying seconds, it’s easier to read.

Also, does anything speak against showing this for just 1s? I’d like to avoid situations where the screen doesn’t change for so long that people think everything is okay and just walk away, not paying much attention to the error message.

unlock_indicator.h
```diff
@@ -14,7 +14,8 @@ typedef enum {
     STATE_PAM_IDLE = 0,   /* no PAM interaction at the moment */
     STATE_PAM_VERIFY = 1, /* currently verifying the password via PAM */
     STATE_PAM_LOCK = 2,   /* currently locking the screen */
-    STATE_PAM_WRONG = 3   /* the password was wrong */
+    STATE_PAM_WRONG = 3,  /* the password was wrong */
+    STATE_PAM_FAIL = 4    /* i3lock failed to start */
```
stapelberg: Please rename STATE_PAM_FAIL to STATE_I3LOCK_LOCK_FAILED, this has nothing to do with PAM :)

eplanet: Good catch, I just copy-pasted without thinking enough about it…

xcb.c
eplanet: Got you for `sleep`, it makes sense. For the sleep time, 2 sec was just a proposition, not to be displayed in a flash. But 1 sec is enough anyway. :-)

PR #105, opened 2016-12-06
Provide option to hide the PAM state text on the unlock indicator

alanctkc:
I found myself really wanting to customize the text (default sans-serif looks mediocre on my machine).

However, in lieu of wanting to bring in pango or overcomplicate the text rendering in some other way, I found I'm also pretty happy with just killing the text altogether. But, I want to be able to keep the unlock indicator around to have some kind of feedback still.

Common enough case for upstream or am I the only one? :wink:

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -70,6 +70,10 @@ show you the current PAM state (whether your password is currently being
 verified or whether it is wrong).
 
 .TP
+.B \-x, \-\-no-text
+Hide the PAM state text on the unlock indicator.
+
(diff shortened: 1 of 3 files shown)


stapelberg:
Sorry, we don’t want to clutter i3lock with options with regards to its visual appearance. I realize this is quite a simple patch, but we’re taking a strict stand on this subject.

alanctkc:
100% understandable! Thanks for the quick response.

PR #107, opened 2016-12-26
Change input slices to be exactly pi/3 in size instead of slightly more

Bownairo:
This changes the way separator lines are drawn on input slices. Now the separator lines overwrite the slice on both sides so that the slice is exactly π/3 in size.

As it is now, slices are (131 π)/384 in size because the less clockwise separator is drawn overwriting part of the slice, and the more clockwise separator is drawn past the end of the slice, extending it by π/128.

### DIFF
--- a/unlock_indicator.c
+++ b/unlock_indicator.c
@@ -296,8 +296,8 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
                       BUTTON_CENTER /* x */,
                       BUTTON_CENTER /* y */,
                       BUTTON_RADIUS /* radius */,
-                      highlight_start + (M_PI / 3.0) /* start */,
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for the pull request! For my education, could you please attach a before/after screenshot?

Bownairo:
Of course, I've attached before and afters of a single slice, and of slices next to each other so that the difference can be seen. Notice that there is overlap and one of the slices is comparably smaller because of it.

Before
before
After
after
Before Full
before-full
After Full
after-full

stapelberg:
I never noticed that :). Thanks!

8carlosf:
Maths! nice of you to spot it, and figuring out what was wrong.. good job sir

PR #108, opened 2016-12-27
Add support for a crash handler

sandsmark:
Let user set a command to be run in case of unexpected program
termination; either a crash or the application suddenly exiting for some
other reason.

This command could e. g. log the user out cleanly with something like:
    dbus-send --system --print-reply --dest=org.freedesktop.login1 \
    /org/freedesktop/login1 'org.freedesktop.login1.Manager.TerminateSession' \
    string:c2

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -28,6 +28,8 @@ i3lock \- improved screen locker
 .RB [\|\-u\|]
 .RB [\|\-e\|]
 .RB [\|\-f\|]
+.RB [\|\--command-on-crash=
+.IR crash-handler.sh \|]
 
 .SH DESCRIPTION
 .B i3lock
@@ -109,6 +111,25 @@ Show the number of failed attempts, if any.
 .B \-\-debug
 Enables debug logging.
 Note, that this will log the password used for authentication to stdout.
+
+.TP
+.BI \fB\-\-command-on-crash= path
+Runs the specified command on unexpected process termination (e.g. crash).
+
+For example to shut down the machine:
+
+.Vb 6
+\& i3lock --command-on-crash=/usr/bin/poweroff
+
+To make logind terminate the session to log you out, put this in
+/usr/local/bin/i3lock-crash.sh and make it executable, then use
+--command-on-crash=/usr/local/bin/i3lock-crash.sh:
+
+.Vb 6
+\& #!/bin/sh
+\& dbus-send --system --print-reply --dest=org.freedesktop.login1 \\
+\&   /org/freedesktop/login1 'org.freedesktop.login1.Manager.TerminateSession' \\
+\&   string:c2
 
 .SH DPMS
 
--- a/i3lock.c
+++ b/i3lock.c
@@ -81,6 +81,35 @@ bool tile = false;
 bool ignore_empty_password = false;
 bool skip_repeated_empty_password = false;
 
+static bool expected_exit = false;
+static char *crash_handler = NULL;
+
+static void exit_handler() {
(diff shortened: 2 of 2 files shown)


i3lock.1
```diff
@@ -110,6 +112,19 @@ Show the number of failed attempts, if any.
 Enables debug logging.
 Note, that this will log the password used for authentication to stdout.
 
+.TP
+.BI \-C\  path \fR,\ \fB\-\-crash-handler= path
+Runs the specified command on unexpected process termination (e. g. crash).
```
Airblader: nit: s/e. g./e.g./

i3lock.1
```diff
@@ -110,6 +112,19 @@ Show the number of failed attempts, if any.
...
+
+For example to log out cleanly:
```
Airblader: Examples are nice, but this one is extremely specific to the system you're using and will not work for many people. I fear it might cause more questions than it does good. Using a systemd command for a forced shutdown might be more widely applicable and also more concise.

i3lock.c
```diff
@@ -802,14 +834,15 @@ int main(int argc, char *argv[]) {
         {"ignore-empty-password", no_argument, NULL, 'e'},
         {"inactivity-timeout", required_argument, NULL, 'I'},
         {"show-failed-attempts", no_argument, NULL, 'f'},
+        {"crash-handler", required_argument, NULL, 'C'},
```
Airblader: I don't think we need a one letter parameter for this. C in particular doesn't really speak to me. Plus it has a particular meaning for other tools from the i3 suite. I think the flag should communicate more clearly what format the argument has (a command). Something like `--command-on-crash`?

i3lock.1
sandsmark: I added an example for using just `poweroff` which should be more universal.

i3lock.c
sandsmark: fixed.

i3lock.1
sandsmark: fixed

stapelberg:
Thanks for your work! I’m currently leaning towards declining this pull request:

* I don’t see any evidence that i3lock is prone to crashes (i.e. I can’t remember any crashes being reported in the recent past).
* If i3lock indeed is prone to crashes, I think we should spend our time/lines of code/maintenance efforts on making the code more robust instead of making the error handling better. I can’t help it: the proposed crash handler reminds me of programs whose main function catches all exceptions and tries to continue. Both cases feel unclean to me.
* Even if we wanted to have better error handling, I’d argue that such handling should happen outside of the screen locker itself and treat the screen locker as a black box. That way, _all_ possible reasons for termination are caught, not just the ones in which the process is still healthy enough to realize it’s about to crash and react to that.

Can you outline why it’s not possible to have such a handler outside of i3lock?

sandsmark:
> Can you outline why it’s not possible to have such a handler outside of i3lock?

Because an outside process doesn't know why the i3lock process is exiting, e. g. if one of the libraries i3 uses calls exit() for some reason.

stapelberg:
> Because an outside process doesn't know why the i3lock process is exiting, e. g. if one of the libraries i3 uses calls exit() for some reason.

I trust the code quality of i3lock’s dependencies to be high enough that there are no stray `exit(0)` calls in there. Hence, you can use the exit code to distinguish success from failure.

sandsmark:
Sure, I was just reminded of the CVEs for kscreenlocker and xscreensaver and got a bit paranoid.

PR #110, opened 2017-01-18
Command line option to center image

XDjackieXD:
This PR add a command line option to display the background image centered on every connected screen.
This allows for some fancy things like displaying a round logo inside the unlock circle and whatever you can come up with.
I chose to call the option "middle" [-m], because [-c] was already taken by "color".

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -23,6 +23,7 @@ i3lock \- improved screen locker
 .RB [\|\-c
 .IR color \|]
 .RB [\|\-t\|]
+.RB [\|\-m\|]
 .RB [\|\-p
 .IR pointer\|]
 .RB [\|\-u\|]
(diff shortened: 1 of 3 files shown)


stapelberg:
Thanks for your pull request. I’ll have to decline merging it, though. Quoting from PR #92:

> We don't want this kind of feature in i3lock because we want to keep it simple and easy to avoid security issues, so unfortunately I'll have to reject and close this PR. There's been many pull requests for the same and similar features where we explain this in more detail. We think this kind of processing is best placed outside of i3 with dedicated graphics software.

stapelberg:
(The guidelines for contributing also mention this: https://github.com/i3/i3lock/blob/master/.github/CONTRIBUTING.md)

XDjackieXD:
oh sorry. I didn't notice the contributing file in the .github folder...
the problem is that with external tools it is possible to center the image by making it the same resolution than the screen. this doesn't help though when using a laptop with external screens or projectors of various resolutions.

stapelberg:
You can use `xrandr -q` to obtain the screen configuration in a wrapper script around i3lock.

XDjackieXD:
using xrandr and creating an image before locking would be a very hacky solution that would be much slower though

stapelberg:
You can cache the result, or even pre-compute it on display changes if that’s important to you.

PR #117, opened 2017-02-25
Add shell exec param on wrong password

ikarulus:
This PR adds the functionality to execute shell commands with the -s param, what is pretty cool when you want to do things like taking a photo with the webcam to identify the person which tried to access your computer.

Hacked together with @luciaagullo and @koraa at @jugendhackt


Airblader:
duplicates #61

PR #122, opened 2017-04-14
Remove last traces of DPMS

jasperla:
Since DPMS support was removed in 2.8, these last bits shouldn't be needed anymore.


jasperla:
Thanks for the prompt merge!

PR #123, opened 2017-04-15
Add support for bsd_auth(3) as authentication backend

jasperla:
This PR adds support for `bsd_auth(3)` as an authentication backend to i3lock. To keep things simple it's a compile-time option and enabled by default on OpenBSD-only.

In the process of keeping the actual addition of `bsd_auth(3)` as small as possible I renamed a lot of non-PAM specific variables/fields that had `PAM/pam` in them to `AUTH/auth`. I think this will also make it easier for future backends to be added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 TOPDIR=$(shell pwd)
+UNAME=$(shell uname)
 
 INSTALL=install
 PREFIX=/usr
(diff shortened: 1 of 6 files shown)


Airblader:
@stapelberg What do you think about this?

@jasperla Generally speaking we don't want compile-time flags in i3lock & family. It should just be platform-specific like the other things, though of course this leaves the pam dependency in OpenBSD. Besides, is there any reason why i3lock on OpenBSD should no longer support pam?

jasperla:
@Airblader it's not so much that i3lock on OpenBSD should no longer support PAM; rather that OpenBSD never had PAM. There is an openpam port but it's merely a stub library than anything else. I can see why you generally wouldn't want any compile time options, but for such fundamental things as the authentication backend that is system-specific (bsd_auth and more or less PAM too) I think it would be justified (I may be biased ;-)).

Airblader:
Thanks for the explanation – you can tell I'm not knowledgable on BSDs. :-) I'd rather leave this decision up to @stapelberg since it's pretty fundamental, as you pointed out yourself.

stapelberg:
I don’t see why this needs to be optional. If OpenBSD doesn’t provide a working PAM stack, i3lock should always use `bsd_auth(3)` on OpenBSD.

Can you update the PR to use ifdefs please?

jasperla:
The code is already using `#ifdef` to conditionally compile either codepath.

stapelberg:
Yes, but the ifdefs aren’t consistent with the way we’re using them in i3status/i3: you’re currently declaring a custom definition (`USE_BSDAUTH` and `USE_PAM`) whereas the rest of the code directly checks under which platform it is compiled.

jasperla:
Oh right, so `#ifdef __OpenBSD__, #else, #endif` ?

stapelberg:
Yes.

jasperla:
That should do it.

stapelberg:
Thanks!

jasperla:
Great, thanks a lot for the merge!

PR #125, opened 2017-04-17
Restore intended behaviour and don't use mlock(2) on OpenBSD.

jasperla:
This addresses an issue that wasn't visible in my earlier testing due to the fact I had `i3lock` setuid root (to be able to call the bsdauth helpers), only to realize setgid auth suffices.

As a side note, i3lock is now packaged in [OpenBSD too](http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/x11/i3lock/Makefile).

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -919,12 +919,12 @@ int main(int argc, char *argv[]) {
         errx(EXIT_FAILURE, "PAM: %s", pam_strerror(pam_handle, ret));
 #endif
 
-/* Using mlock() as non-super-user seems only possible in Linux and OpenBSD.
+/* Using mlock() as non-super-user seems only possible in Linux.
  * Users of other operating systems should use encrypted swap/no swap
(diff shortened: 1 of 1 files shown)


Airblader:
Sorry that this took a while, I must've accidentally deleted the email about it. Thanks! Great to hear it's packaged now!

jasperla:
Not a problem at all, thanks for the merge.

jasperla:
@stapelberg @Airblader  would it be possible to roll a new release sometime soon-ish includes the recent ``bsdauth`` additions?

Airblader:
@stapelberg Yeas, a release makes sense both timewise and with the number of changes, I think. The release for i3lock isn't scripted, so I don't know if you can do it or how I can help?

stapelberg:
Sure. I’m currently somewhat busy, but I expect to be able to do a release within the next 2-3 weeks.

PR #126, opened 2017-04-25
Alteration of login failed message and font size

vishwajithsandaru:
"Wrong" message for an incorrect password is changed to "access denied",  and the font size is reduced.

### DIFF
--- a/unlock_indicator.c
+++ b/unlock_indicator.c
@@ -204,7 +204,7 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
 
         cairo_set_source_rgb(ctx, 0, 0, 0);
         cairo_select_font_face(ctx, "sans-serif", CAIRO_FONT_SLANT_NORMAL, CAIRO_FONT_WEIGHT_NORMAL);
-        cairo_set_font_size(ctx, 28.0);
+        cairo_set_font_size(ctx, 20.0);
         switch (auth_state) {
             case STATE_AUTH_VERIFY:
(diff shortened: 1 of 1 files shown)


Airblader:
Putting aside that the commits are broken and include two independent changes... What are the reasons for these changes?

vishwajithsandaru:
In practical its better to say "access  denied" than "wrong" during a login failure to a computer, but when the message is changed to "access denied" the default font size seems to be too large and there is a lack of space to some of the letters and they are going out of the field, so I reduced the font size littlebit.

Airblader:
> In practical its better to say "access denied" than "wrong" 

That's just restating your change, it doesn't give an argument. We currently return the information PAM gives us, I see no reason to change this.

vishwajithsandaru:
Sorry for the inconvenience , actually I  stated a common problem that arose from user experiences in my area. If you are able could you please explain me the thing with the authentication module - "We currently return the information PAM gives us" ...

Airblader:
> If you are able could you please explain me the thing with the authentication module - "We currently return the information PAM gives us" ...

That was actually a mistake on my side. :-) It's not the PAM result, just our internal status. Still, I feel that "Access Denied", as it stands, is inconsistent with the other texts we display ("locking…", "verifying…", "lock failed!") and I don't see a convincing reason for any change other than personal preference.

Airblader:
I'm going to close this for now for the mentioned reasons.

PR #136, opened 2017-07-09
Only use -lpam when not on OpenBSD

kaashif:
There was a regression in 10416e27, the Makefile uses -lpam unconditionally. This causes the build to fail on OpenBSD. Deleting the line fixes it since -lpam is later added to LIBS if we're not on OpenBSD.


Airblader:
Thanks!

PR #137, opened 2017-07-09
Remove unconditional -lpam that was accidentally added back in 10416e27

jasperla:


Airblader:
Thanks! I already merged #136, so we can ignore this PR.

jasperla:
Thanks!

PR #140, opened 2017-08-13
Change the locale discovery procedure to treat empty string same as unset

bluetech:
This is explained in the commit message here:
https://github.com/xkbcommon/libxkbcommon/commit/f468f0b2430e15cc262c5745445580bd0dc64ef0


Airblader:
Thanks!

PR #153, opened 2017-10-21
Measure wall-clock time instead of CPU time for “locking” indicator.

stapelberg:
related to https://github.com/i3/i3lock/issues/35

Changes the magic number to 100ms, which is commonly believed to be the threshold at which humans perceive changes as “instantaneous”. This is large enough to prevent flickering, and small enough to fix the issue.

### DIFF
--- a/xcb.c
+++ b/xcb.c
@@ -19,6 +19,7 @@
 #include <assert.h>
 #include <err.h>
 #include <time.h>
+#include <sys/time.h>
 
 #include "cursors.h"
 #include "unlock_indicator.h"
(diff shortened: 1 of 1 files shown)


stapelberg:
Travis fails until this PR is rebased on top of https://github.com/i3/i3lock/pull/154

xcb.c
```diff
@@ -199,10 +202,18 @@ void grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
...
+        }
+
+        struct timeval elapsed;
+        timersub(&now, &start, &elapsed);
+
+        /* Trigger a screen redraw if 100ms elapsed */
         if (!redrawn &&
             (tries % 100) == 0 &&
-            (clock() - start) > 250000) {
+            elapsed.tv_usec > 100000) {
```
Airblader: Is there a specific reason to change the timing itself? Also, why don't we consider using much, much fewer grabbing retries to cut down the time? xscreensaver only uses a handful. This would give much quicker feedback.

xcb.c
```diff
@@ -226,10 +237,18 @@ void grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
...
```
Airblader: Since this number is used twice, it should perhaps be extracted and not be a magic number.

bebehei:
Thank you @stapelberg for taking care about this.

xcb.c
stapelberg: See the PR description:

> Changes the magic number to 100ms, which is commonly believed to be the threshold at which humans perceive changes as “instantaneous”. This is large enough to prevent flickering, and small enough to fix the issue.

We can also cut down the number of retries, but that seems orthogonal: the number of retries determines how long i3lock can hang, not how quickly the user is presented with feedback — that would still be coupled to wall-clock time measurement, as retries are not deterministic in length either (just like `clock`, which we’re removing here).

xcb.c
stapelberg: Done.

PR #155, opened 2017-10-22
SetInputFocus to the i3lock window to force-close context menus

stapelberg:
When grabbing the pointer/keyboard fails, a new code path is activated, which:

1. Uses the standards-compliant _NET_ACTIVE_WINDOW root window property to
   determine the window to restore focus to.

2. Sets the input focus to the i3lock window, effectively force-closing any open
   context menus (tested with Google Chrome).

3. Upon exiting, restores focus to the window from step ① by sending a
   _NET_ACTIVE_WINDOW ClientMessage to the root window. Note that this step
   requires https://github.com/i3/i3/pull/3027 in i3 to not mess up focus.

fixes https://github.com/i3/i3lock/issues/35

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -35,6 +35,7 @@
 #ifdef __OpenBSD__
 #include <strings.h> /* explicit_bzero(3) */
 #endif
+#include <xcb/xcb_aux.h>
 
 #include "i3lock.h"
 #include "xcb.h"
@@ -278,7 +279,8 @@ static void input_done(void) {
         DEBUG("successfully authenticated\n");
         clear_password_memory();
 
-        exit(0);
+        ev_break(EV_DEFAULT, EVBREAK_ALL);
+        return;
     }
 #else
     if (pam_authenticate(pam_handle, 0) == PAM_SUCCESS) {
@@ -292,7 +294,8 @@ static void input_done(void) {
         pam_setcred(pam_handle, PAM_REFRESH_CRED);
         pam_end(pam_handle, PAM_SUCCESS);
 
-        exit(0);
+        ev_break(EV_DEFAULT, EVBREAK_ALL);
+        return;
     }
 #endif
 
@@ -1010,6 +1013,8 @@ int main(int argc, char *argv[]) {
     /* Pixmap on which the image is rendered to (if any) */
     xcb_pixmap_t bg_pixmap = draw_image(last_resolution);
 
+    xcb_window_t stolen_focus = find_focused_window(conn, screen->root);
+
     /* Open the fullscreen window, already with the correct pixmap in place */
     win = open_fullscreen_window(conn, screen, color, bg_pixmap);
     xcb_free_pixmap(conn, bg_pixmap);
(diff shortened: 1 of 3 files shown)


i3lock.c
```diff
@@ -1018,7 +1023,24 @@ int main(int argc, char *argv[]) {
...
-    grab_pointer_and_keyboard(conn, screen, cursor);
+    if (!grab_pointer_and_keyboard(conn, screen, cursor, 1000)) {
+        stolen_focus = find_focused_window(conn, screen->root);
+        DEBUG("stole focus from X11 window 0x%08x\n", stolen_focus);
+
+        /* Set the focus to i3lock, possibly closing context menus which would
+         * otherwise prevent us from grabbing keyboard/pointer.
+         *
+         * We cannot use set_focused_window because _NET_ACTIVE_WINDOW only
+         * works for managed windows, but i3 uses an unmanaged window
```
Airblader: i3lock, not i3

i3lock.c
```diff
@@ -1065,4 +1087,17 @@ int main(int argc, char *argv[]) {
...
+
+    if (stolen_focus == XCB_NONE) {
+        return 0;
+    }
+
+    DEBUG("restoring focus to X11 window 0x%08x\n", stolen_focus);
+    xcb_ungrab_pointer(conn, XCB_CURRENT_TIME);
+    xcb_ungrab_keyboard(conn, XCB_CURRENT_TIME);
+    xcb_destroy_window(conn, win);
+    set_focused_window(conn, screen->root, stolen_focus);
```
Airblader: Why don't we use SetInputFocus here as well? We're already using it above. What if prior to locking the focus belonged to an unmanaged window? We are simply restoring focus to the wrong window here.

i3lock.c
stapelberg: Done.

i3lock.c
stapelberg: > Why don't we use SetInputFocus here as well? We're already using it above.

Because _NET_ACTIVE_WINDOW is the better option: it is a request to the window manager, whereas SetInputFocus circumvents the window manager. The fact that we must resort to SetInputFocus for the part where focus concerns an unmanaged window doesn’t mean we have to give up on _NET_ACTIVE_WINDOW in the other cases.

> What if prior to locking the focus belonged to an unmanaged window? We are simply restoring focus to the wrong window here.

If an unmanaged window was focused, a different window will be focused, yes. i3lock, without my proposed change, already screws up focus when the pointer is over a different window, so we’re not making things worse: we’re fixing a somewhat common case at the expense of a much rarer case:

Unmanaged windows are quite the edge case. Given that i3 in general prioritizes tiling windows and considers floating windows a nuisance, I think it is fair to say that we don’t need to care about focus restore behavior for unmanaged windows when unlocking your screen.

Airblader:
> What are you referring to? On my machine, this method was reliable in 100% of my tests.

As I've pointed out in the issue, for example for thunar's context menu this doesn't work. The truth is, no one has actually put any work into figuring out whether this approach works in most, some or almost no cases. We only know for sure that it's neither 0% nor 100%. 

In the issue I event went as far as saying that by making it less transparent when i3lock does or doesn't work, we might just induce faulty habits in users:

> *Users now might just learn the habit of leaving context menus open because »it works« – until it doesn't.*

---

> Because _NET_ACTIVE_WINDOW is the better option: it is a request to the window manager, whereas SetInputFocus circumvents the window manager.

After having bypassed the window manager before already, why is this better *now*? What is the benefit of going through the window manager, specifically? I can't really think of a good reason.

However, on the contrary, I could name the following contra arguments:

* Using EWMH increases i3lock's code complexity.
* Restoring focus will break for non-EWMH compliant window managers (although, granted, a window manager not supporting _NET_ACTIVE_WINDOW isn't to be taken too seriously these days)
* This change forces us to teach i3 how to deal with this scenario. This wasn't necessary until now, suggesting that i3lock is implementing something quite unusual here.

> Unmanaged windows are quite the edge case.

Context menus are unmanaged windows as well, but that doesn't seem to stop us here.

Airblader:
And just for the fun of it, I've also tried it with gedit where this approach fails as well.

stapelberg:
> And just for the fun of it, I've also tried it with gedit where this approach fails as well.

Updated the commit description to reflect the examples of working/not working (not aiming for an exhaustive list).

> After having bypassed the window manager before already, why is this better now? What is the benefit of going through the window manager, specifically?

Going through the window manager is the correct way to ensure that the _NET_ACTIVE_WINDOW window on the root property is updated. Even though the value stays the same, the notification serves to wake up clients which react to losing focus. Think of an instant messenger which produces message read receipts, for example.

When using SetInputFocus directly, we’d need to also update _NET_ACTIVE_WINDOW, which seems like the less clean solution of the two to me.

Airblader:
So then we have to update the property for when we steal focus since there we aren't using the window manager, no?

stapelberg:
> So then we have to update the property for when we steal focus since there we aren't using the window manager, no?

My understanding is that no, the _NET_ACTIVE_WINDOW property should only be pointing to managed windows, and we are stealing focus to an unmanaged window, so no update is required.

Airblader:
Ah, yeah, right. :-)

stapelberg:
Duly noted :). Maybe a better way presents itself eventually and makes us all happy. For now, let’s go ahead with this.

Issue #68, opened 2016-04-11
JPEG support


boughtonp:
I want to use jpg files on my lock screen.

Is JPEG support deliberately excluded, or simply not included because it doesn't come with Cairo?

It looks like there's [cairo_jpg](https://github.com/rahra/cairo_jpg) which would provide the necessary functionality.

Is that something that could be added to i3lock? (Or is there a fork which supports JPEGs?)

Airblader:
Is there a specific reason not to use something like imagemagick to just convert it to png?

boughtonp:
That means duplicating a file every time the image is changed.

stapelberg:
jpeg is excluded because it doesn’t come with cairo. I’m not keen on adding a new dependency just to support jpeg.

lindhe:
I get that the target audience for i3lock might not be your average Joe, but jpeg is (based on statistics I made up, but seems reasonable) more common than png. Especially for photos. So it's a rather common case to have a jpg laying around that you want to use as a background.

Sure you *can* convert it, but is it reasonable to ask that from the user? Maybe. Would it be more convenient for the user if the program supported jpg naively? Definitely.

Also, it does not give a nice error message when loading with a jpg. For me it just says `Could not load image "Pictures/images/wallpaper.jpg": out of memory`. Should probably say "Can only load PNG images.".

stapelberg:
See #114 for details about the error message; I agree that should be improved.

I also hear what you’re saying, but the answer remains that the arguments don’t outweigh the cost of an additional dependency.

lucianoq:
@Airblader Actually I found some specific reasons.

```
$ time -p scrot -t 10 qq.png
real 0.53
user 0.48
sys 0.01

$ time -p scrot -t 10 qq.jpg
real 0.11
user 0.06
sys 0.02

$ time -p convert qq-thumb.jpg -resize 1000% qq.png
real 1.68
user 2.05
sys 0.11

$ time -p convert qq-thumb.jpg -resize 1000% qq.jpg
real 0.18
user 0.57
sys 0.11

$ time -p convert qq-thumb.png -resize 1000% qq.png
real 1.53
user 1.93
sys 0.15

$ time -p convert qq-thumb.png -resize 1000% qq.jpg
real 0.24
user 0.65
sys 0.18
```
Or, aggregated:
```
$ time -p sh -c "scrot -t 10 qq.jpg && convert qq-thumb.jpg -resize 1000 qq.jpg"
real 0.14
[...]

PR #157, opened 2017-10-31
Splitting image loading out of the main file, and adding JPEG support.

### LINKED ISSUE #114: Providing an invalid PNG gives confusing error message "out of memory"
I saw in another issue the rationale for no JPG support or what have you, but it seems a little cryptic to give an out of memory error message when the the file is not a valid PNG.

## How to reproduce

```
➜  ~  wget -q https://upload.wikimedia.org/wikipedia/commons/d/d6/Bromo-Tengger-Semeru_banner_2.jpg
➜  ~  i3lock -i Bromo-Tengger-Semeru_banner_2.jpg
Could not load image "Bromo-Tengger-Semeru_banner_2.jpg": out of memory
➜  ~  i3lock -i /dev/zero                        
Could not load image "/dev/zero": out of memory
```

Cynary:
See #68 for discussion and context.

Also did some cleanup:
- Repeated string.h header include in i3lock.c;
- We were not destroying the cairo surface;

Also revamped error management during image loading. We won't get out of memory errors for
invalid types anymore. Fixes #114 

Finally, added libjpeg dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
     - libev-dev
     - libxcb-xinerama0-dev
     - libxcb-xkb-dev
+    - libjpeg-dev
 before_install:
   - "echo 'APT::Default-Release \"trusty\";' | sudo tee /etc/apt/apt.conf.d/default-release"
   - "echo 'deb http://archive.ubuntu.com/ubuntu/ xenial main universe' | sudo tee /etc/apt/sources.list.d/wily.list"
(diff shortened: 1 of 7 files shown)


stapelberg:
Thanks for sending this PR.

The complexity of the JPEG-related code confirms my intuition that I don’t feel comfortable maintaining this in the context of i3lock. I think this should either go into cairo upstream or into a third-party library which would ideally be used by many, not just i3lock.

I’d still be interested in the remaining cleanups you did, so if you’d like to submit these as a separate PR, that’d be appreciated!

PR #158, opened 2017-11-05
add the regretable timer feature

tsankuanglee:
Give user an initial time window to cancel the lock by any key.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -18,6 +18,8 @@ i3lock \- improved screen locker
 .RB [\|\-v\|]
 .RB [\|\-n\|]
 .RB [\|\-b\|]
+.RB [\|\-r
+.IR timeout \|]
 .RB [\|\-i
 .IR image.png \|]
 .RB [\|\-c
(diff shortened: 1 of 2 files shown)


stapelberg:
Thanks for the PR.

Unfortunately, I’ll have to decline merging this: I think the potential for abuse is too big to justify the benefit of the feature: an attacker could enable the feature, wait until the victim locks their screen and walks away, then conveniently unlock the screen.

We’ve had multiple reports in the past where even a split-second impression of a locked screen was enough to trick people into thinking their screen was locked when it really wasn’t.

tsankuanglee:
Not sure why this feature is more insecure than an already compromised machine on which an attacker could change things.

The feature is the same as XScreenSaver's Fade Seconds feature. Perhaps this feature should be left for the screensaver, not the screen locker?

Anyway, thanks for the project.

stapelberg:
> Not sure why this feature is more insecure than an already compromised machine on which an attacker could change things.

Machines aren’t necessarily in a clear compromised or non-compromised state. It’s too easy to sneak an additional argument into someone’s i3lock invocation, which opens a much bigger threat.

> The feature is the same as XScreenSaver's Fade Seconds feature.

A crucial distinction that I see between the two is that AFAIK, XScreenSaver will provide visual feedback that a fade is going on.

> Perhaps this feature should be left for the screensaver, not the screen locker?

That actually sounds best to me :)

PR #163, opened 2017-11-25
Switch to autotools

stapelberg:
This was largely copied from the i3 configure.ac and Makefile.am.

cc @jasperla @kaashif — I haven’t tested this on non-Linux, so if you could test this, that’d be appreciated.


Airblader:
Obviously not going to review all the details of this. :-)

stapelberg:
I’m going to go ahead and merge this. Feedback for non-linux still very welcome, but now in form of a pull request please :).

jasperla:
This mostly looks fine on OpenBSD, except for two small issues:
- it explicitly checks for `iconv_open()` which fails. it seems there is [upstream documentation](https://www.gnu.org/software/gettext/manual/html_node/AM_005fICONV.html) on the topic. Perhaps i3lock (and i3 too for that matter) should use the more portable approach of `AM_ICONV()`?
- pam is unconditionally linked in (pull request forth-coming)

stapelberg:
Thanks for looking. Pull requests for both issues welcome :)

PR #165, opened 2017-12-02
make the check for libpam conditional

jasperla:
After the switch to autoconf i3lock links unconditionally against libpam. This PR makes this optional but defaults to the original behaviour of checking.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -76,7 +76,15 @@ AC_SEARCH_LIBS([ev_run], [ev], , [AC_MSG_FAILURE([cannot find the required ev_ru
 
 AC_SEARCH_LIBS([shm_open], [rt])
 
-AC_SEARCH_LIBS([pam_authenticate], [pam])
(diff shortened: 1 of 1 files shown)


stapelberg:
Given that we don’t have optional features in i3lock (unless the platform is lacking it), could we tie this feature to platform support please? I.e., auto-disable pam on OpenBSD, auto-enable elsewhere.

jasperla:
The proposed patch is to auto-enable it anywhere, only on OpenBSD one has to pass `--disable-pam`. But if that isn't the preferred approach I can change it to look at the OS-triplet to determine if it needs to check for PAM or skip the check.

stapelberg:
> I can change it to look at the OS-triplet to determine if it needs to check for PAM or skip the check.

Yes, please change it to look at the OS triplet.

jasperla:
Updated to look at `$host`.

Issue #167, opened 2017-12-08
memleak in cairo_select_font_face


PandorasFox:
I've squashed all the memleaks in my fork so far except for the small handful that are in libfontconfig.so, which are also . They all seem to be coming from the calls to [`cairo_select_font_face`](https://www.cairographics.org/manual/cairo-text.html#cairo-select-font-face), which is still apparently just a 'toy' function.

I'll be poking at using cairo glyphs for the text instead, which should be a bit more robust (and not have memory leaks), and should have a fix soonish, if that's alright.

Airblader:
I'm sorry, I'm a bit confused. What's the issue in i3lock here?

PandorasFox:
ASAN complains some about memory errors in fontconfig due to cairo's `show_text` and `select_font_face` functions being toy/non-production functions. (I think that with a regular version of fontconfig, only ~3 memory leaks are reported, but with a version compiled with ASAN there's... A few thousand).

Additionally, since i3lock keeps setting the font face through cairo's toy functions on each redraw, it's a bit inefficient and hits up fontconfig on each tick. Depending on the version of fontconfig at runtime, this might cause a memory error on each tick.

Airblader:
Thanks for the explanation; so the i3lock issue here is the unnecessary calls, right? The memory errors seem to be an issue of cairo if I understand you correctly. :-)

PandorasFox:
Pretty much. The real issue with i3lock is using the "toy" Cairo functions, which cause the memory leaks.

I'm awful at naming things, including issues :)

Airblader:
I haven't looked into those functions, but if they don't do anything we can also remove the calls.

PandorasFox:
They're currently necessary for displaying the "Verifying..." /"Incorrect" /etc text, but they're basically the quick/inefficient way to get text done with Cairo.

The more correct way is for i3lock to lookup the font, then supply it to Cairo and cache that, to avoid using the toy functions _and_ repeated lookups, since the toy functions don't include reference counting on the fonts they return/don't guarantee longevity.

Airblader:
If the efforts to do so are reasonable, and I assume they are, I think it'd make sense to change it to what you described. PRs welcome, but I think you already mentioned you're preparing for one, right? :-)

PandorasFox:
I made #168 last night :)

feddasch:
Is #168 going to be merged soon?

psychon:
[...]

PR #168, opened 2017-12-09
Move to use cairo glyphs instead of text

PandorasFox:
Requires linking fontconfig now, but that was already required at runtime for cairo anyways.

Fixes the memory leaks caused by using  [`cairo_select_font_face`](https://www.cairographics.org/manual/cairo-text.html#cairo-select-font-face), as mentioned in #167


unlock_indicator.c
```diff
@@ -203,63 +283,54 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
...
+                            text = buf;
+                        }
+                        cairo_set_source_rgb(ctx, 1, 0, 0);
+                        cairo_set_font_size(ctx, 32.0);
+                    }
+                    break;
+            }
 
-            cairo_set_font_size(ctx, 14.0);
+            if (text ) {
```
Airblader: Broken formatting

Airblader:
@stapelberg I can offer a review in terms of checking the code, but given the changes to the dependencies I'd assume you would want to look at this, too.

unlock_indicator.c
```diff
@@ -91,6 +94,83 @@ static double scaling_factor(void) {
...
+        DEBUG("no sans-serif font available\n");
+        return NULL;
+    }
+
+    FcDefaultSubstitute(pattern);
+    if (!FcConfigSubstitute(FcConfigGetCurrent(), pattern, FcMatchPattern)) {
+        DEBUG("config sub failed?\n");
+        return NULL;
+    }
+    FcPattern *tmp = FcFontMatch(FcConfigGetCurrent(), pattern, &result);
```
stapelberg: You can add a `FcPatternDestroy(pattern)` call right below this line to get rid of the duplication below.

unlock_indicator.c
```diff
@@ -91,6 +94,83 @@ static double scaling_factor(void) {
...
+    if (!tmp) {
+        DEBUG("no sans-serif font available\n");
+        FcPatternDestroy(pattern);
+        return NULL;
+    }
+
+    FcPatternDestroy(pattern);
+    pattern = tmp;
+    tmp = NULL;
```
stapelberg: Why not keep using `tmp`? I think that’s clearer than re-purposing a variable.

unlock_indicator.c
```diff
@@ -91,6 +94,83 @@ static double scaling_factor(void) {
...
+ */
+
+static cairo_font_face_t *get_font_face() {
+    if (cairo_font_face_get_reference_count(sans_serif)) {
+        return cairo_font_face_reference(sans_serif);
+    }
+
+    FcResult result;
+    FcBool res = FcInit();
+    if (!res) {
```
stapelberg: Why not just `if (!FcInit())`?

unlock_indicator.c
```diff
@@ -91,6 +94,83 @@ static double scaling_factor(void) {
...
+        DEBUG("Fontconfig init failed. No fonts will be available...\n");
+        return NULL;
+    }
+
+    FcPattern *pattern = FcNameParse(font_face_name);
```
stapelberg: Can you add some comments to the various Fc* function calls here outlining what they do? Also, is this following some sort of example or document? If so, can you add a comment with a link to it please?

stapelberg:
I don’t mind the explicit dependency on fontconfig.

Once we’re happy with the code, we should get someone with cairo knowledge to review. Perhaps psychon?

unlock_indicator.c
PandorasFox: > Also, is this following some sort of example or document?

Sort of. I found some partial examples for looking up fonts/drawing text with them, but none of them properly handled memory cleanup.

PandorasFox:
I'll make those changes tomorrow morning. I think I developed this patch on this laptop (I'm traveling for the holidays), so I should have the test programs I made to figure out how clean up the fonctconfig structs properly.

unlock_indicator.c
PandorasFox: The only time the font_size was set to 14px was [here](https://github.com/i3/i3lock/blob/master/unlock_indicator.c#L252). That got moved to [here](https://github.com/i3/i3lock/pull/168/files/0a03012d5af117e8631c23d66fb575daa0bbc5b0#diff-efe870ebf4bb57bf7e8e1316d4526cdaR327) in this patch.

PandorasFox:
Alright, I added some comments and made the changes with variable handling/duplicate lines.

I couldn't find the files I used for reference when working on this, so I'm assuming they're on my computer back home. I'll keep poking around to see if I can find a gist or something.

unlock_indicator.c
Airblader: Sorry, that's not what I meant. I meant that the code formatting is wrong here because there's an extra space after `text`.

PandorasFox: Ahh, my bad. I didn't notice that, I'll fix that.

unlock_indicator.c
```diff
@@ -91,6 +94,97 @@ static double scaling_factor(void) {
...
+        return cairo_font_face_reference(sans_serif);
+    }
+
+    FcResult result;
+    /*
+     * Loads the default config.
+     * On successive calls, does no work and just returns true.
+     */
+    if (!FcInit()) {
+        DEBUG("Fontconfig init failed. No fonts will be available...\n");
```
stapelberg: What does “no fonts will be available” imply? Will no text be visible on screen? If so, make the message say that :)

unlock_indicator.c
```diff
@@ -91,6 +94,97 @@ static double scaling_factor(void) {
...
+    FcDefaultSubstitute(pattern);
+    if (!FcConfigSubstitute(FcConfigGetCurrent(), pattern, FcMatchPattern)) {
+        DEBUG("config sub failed?\n");
+        return NULL;
+    }
+    /*
+     * Looks up the font pattern and does some internal RenderPrepare work,
+     * then returns the result.
+     */
+    FcPattern *pattern_ready = FcFontMatch(FcConfigGetCurrent(), pattern, &result);
```
stapelberg: result is assigned but not used. Can we either check it or remove it?

unlock_indicator.c
```diff
@@ -91,6 +94,97 @@ static double scaling_factor(void) {
...
+     */
+    FcPattern *pattern = FcNameParse(font_face_name);
+    if (!pattern) {
+        DEBUG("no sans-serif font available\n");
+        return NULL;
+    }
+
+    /*
+     * Gets the default font for our pattern. (Gets the default sans-serif font face)
+     * Required to make the following FcFontMatch call for some FontConfig bookkeeping.
```
stapelberg: I don’t quite understand the last sentence. I think the Fc*Substitute calls are required for the FcFontMatch call to work, but what’s the bit about book-keeping?

Would “Without this step, the FcFontMatch call will fail because…” be an accurate description? If so, could you take it and complete it please?

unlock_indicator.c
```diff
@@ -91,6 +94,97 @@ static double scaling_factor(void) {
...
+    FcPattern *pattern_ready = FcFontMatch(FcConfigGetCurrent(), pattern, &result);
+    FcPatternDestroy(pattern);
+    pattern = NULL;
+    if (!pattern_ready) {
+        DEBUG("no sans-serif font available\n");
+        return NULL;
+    }
+
+    /*
+     * Passes the given pattern into cairo, which loads it into a caito freetype font face.
```
stapelberg: type: caito → cairo

unlock_indicator.c
```diff
@@ -91,6 +94,97 @@ static double scaling_factor(void) {
...
+     * On successive calls, does no work and just returns true.
+     */
+    if (!FcInit()) {
+        DEBUG("Fontconfig init failed. No fonts will be available...\n");
+        return NULL;
+    }
+    /*
+     * converts a font face name to a pattern for that face name
+     */
+    FcPattern *pattern = FcNameParse(font_face_name);
```
stapelberg: As `font_face_name` is only used once, could you just write "sans-serif" here?

unlock_indicator.c
```diff
@@ -91,6 +94,97 @@ static double scaling_factor(void) {
...
+    cairo_text_extents(ctx, text, &extents);
+    x = BUTTON_CENTER - ((extents.width / 2) + extents.x_bearing);
+    y = BUTTON_CENTER - ((extents.height / 2) + extents.y_bearing) + offset;
+
+    status = cairo_scaled_font_text_to_glyphs(cairo_get_scaled_font(ctx),
+            x, y,
+            text, -1,
+            &glyphs, &num_glyphs,
+            NULL, NULL, NULL);
+    if (status == CAIRO_STATUS_SUCCESS) {
```
stapelberg: Let’s inverse this conditional like so:

```c
if (status != CAIRO_STATUS_SUCCESS) {
    DEBUG(…);
    return;
}
cairo_show_glyhps(…);
// …
`

File: unlock_indicator.c [hunk 12]
```diff
@@ -91,6 +94,97 @@ static double scaling_factor(void) {
...
+ * Returns the cairo_font_face_t for sans-serif
+ * if this thing returns NULL, then we're somehow on a system without a
+ * sans-serif font, or a system without fontconfig (but with cairo somehow)
+ * Reference counts and caches the font-face, so we don't have memory leaks
+ * (and so that we don't have to go through all the parsing each time)
+ */
+
+static cairo_font_face_t *get_font_face() {
+    if (cairo_font_face_get_reference_count(sans_serif)) {
+        return cairo_font_face_reference(sans_serif);
```
stapelberg: Is the reference count increase/decrease actually necessary? There’s just one location where we use the font face.

stapelberg:
I’m not entirely sure why the travis check doesn’t complain, but formatting the code with `clang-format-3.5 -i *.[ch]` shows differences for me. Before pushing, could you please run that command to ensure consistent formatting? Thanks!

unlock_indicator.c
PandorasFox: Not really. I can just change it to `if (sans_serif) return sans_serif;` pretty much, I guess.

unlock_indicator.c
PandorasFox: Yeah, I'll just do the `unsigned char*` cast inline, I guess. I moved it out to a variable to make the code read a bit more cleanly.

unlock_indicator.c
PandorasFox: It's pretty much just required to update FontConfig's internal state (I'm still not entirely sure why). I'll change the description.

unlock_indicator.c
PandorasFox: This is a fun one. It has to be assigned (or else fontconfig will error out), but its value on a success isn't well-defined. In [fonctconfig's provided examples](https://github.com/behdad/fontconfig/blob/7ca28c2fedb34c1db5ee3116d335f15195859db0/fc-match/fc-match.c#L213) they also only check to see if they get back `NULL` or not :(

I'll update the commenting to say `returns the resulting pattern` to be a bit more clear, though.

PandorasFox:
> Before pushing, could you please run that command to ensure consistent formatting? Thanks!

Should be good now!

unlock_indicator.c
```diff
@@ -80,6 +80,9 @@ static xcb_visualtype_t *vistype;
 unlock_state_t unlock_state;
 auth_state_t auth_state;
 
+/* Cache the font we use after looking it up once, and reference count it */
```
stapelberg: The reference count bit is no longer accurate; remove please

unlock_indicator.c
```diff
@@ -80,6 +80,9 @@ static xcb_visualtype_t *vistype;
...
+cairo_font_face_t *sans_serif = NULL;
```
stapelberg: This can be marked static, as it’s not used outside of the file, yes?

unlock_indicator.c
```diff
@@ -91,6 +94,99 @@ static double scaling_factor(void) {
...
 }
 
+/*
+ * Returns the cairo_font_face_t for sans-serif
+ * if this thing returns NULL, then we're somehow on a system without a
+ * sans-serif font, or a system without fontconfig (but with cairo somehow)
+ * Reference counts and caches the font-face, so we don't have memory leaks
+ * (and so that we don't have to go through all the parsing each time)
+ */
+
```
stapelberg: nit: remove newline between comment and function definition

unlock_indicator.c
```diff
@@ -91,6 +94,99 @@ static double scaling_factor(void) {
...
+static cairo_font_face_t *get_font_face() {
```
stapelberg: nit: C requires functions taking no parameters to read `static cairo_font_face_t *get_font_face(void)`, otherwise they are interpreted as variadic functions

unlock_indicator.c
```diff
@@ -91,6 +94,99 @@ static double scaling_factor(void) {
...
+     * Gets the default font for our pattern. (Gets the default sans-serif font face)
+     * Without these two calls, the FcFontMatch call will fail due to FcConfigGetCurrent()
+     * not giving it a valid/useful config.
+     */
+    FcDefaultSubstitute(pattern);
+    if (!FcConfigSubstitute(FcConfigGetCurrent(), pattern, FcMatchPattern)) {
+        DEBUG("config sub failed?\n");
+        return NULL;
+    }
+    /*
```
stapelberg: nit: add empty line above for consistency

stapelberg:
@psychon, would you have a minute to review this pull request? AFAIR, you’re somewhat familiar with the cairo APIs :) Thanks in advance!

unlock_indicator.c
```diff
@@ -91,6 +94,99 @@ static double scaling_factor(void) {
...
+
+    cairo_text_extents(ctx, text, &extents);
+    x = BUTTON_CENTER - ((extents.width / 2) + extents.x_bearing);
+    y = BUTTON_CENTER - ((extents.height / 2) + extents.y_bearing) + offset;
+
+    status = cairo_scaled_font_text_to_glyphs(cairo_get_scaled_font(ctx),
+                                              x, y,
+                                              text, -1,
+                                              &glyphs, &num_glyphs,
+                                              NULL, NULL, NULL);
```
psychon: I don't know why `cairo_show_text` is marked as toy API, but `cairo_scaled_font_text_to_glyphs` is not, but this code is (basically) doing what `cairo_show_text` would do (well, `cairo_show_text` uses and updates the current point instead of having an extra parameter for that, but that doesn't really matter, I think). So, I would suggest to keep using `cairo_show_text` here.

unlock_indicator.c
```diff
@@ -91,6 +94,99 @@ static double scaling_factor(void) {
...
+
+    /*
+     * Passes the given pattern into cairo, which loads it into a cairo freetype font face.
+     * Increment its reference count and cache it.
+     */
+    cairo_font_face_t *face = cairo_ft_font_face_create_for_pattern(pattern_ready);
+    FcPatternDestroy(pattern_ready);
+    sans_serif = cairo_font_face_reference(face);
+    return face;
+}
```
psychon: I do not really know "font stuff" well, but why not replace the above function with
```C
if (sans_serif == NULL)
    sans_serif = cairo_toy_font_face_create("sans-serif",  CAIRO_FONT_SLANT_NORMAL, CAIRO_FONT_WEIGHT_NORMAL);
return sans_serif;
```
If "memory leaks" is the reason, then I would rather fix that instead of working around it. I am pretty sure that using the toy font backend should not cause memory leaks. And the reason that the toy font API is marked as toy are AFAIK things like right-to-left text and complicated scripts, not memory leaks.

PandorasFox: Doing that'll cause some memory leaks in `fontconfig`. The only way to avoid them is to make the `cairo_debug_reset_static_data` call, which can't be used in conjunction with our own caching.

It definitely seems like it's better to just let i3lock handle the fairly simple lookup + caching than to have to flush the static data every draw tick.

PandorasFox: These are the specific memory leaks I get from ASAN:

```=================================================================
==15286==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x7fbb92b01ae9 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:62
    #1 0x7fbb912287ee  (/usr/lib/libfontconfig.so.1+0x1d7ee)

Indirect leak of 96 byte(s) in 3 object(s) allocated from:
    #0 0x7fbb92b01ce1 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:70
    #1 0x7fbb91228ed9  (/usr/lib/libfontconfig.so.1+0x1ded9)

Indirect leak of 11 byte(s) in 1 object(s) allocated from:
    #0 0x7fbb92a983b1 in __interceptor_strdup /build/gcc/src/gcc/libsanitizer/asan/asan_interceptors.cc:560
    #1 0x7fbb912281f5 in FcValueSave (/usr/lib/libfontconfig.so.1+0x1d1f5)

```

psychon: Hm, okay, that's not the most useful backtrace in the world. Still, I would have expected way more leak reports. When I tried to get awesome valgrind-clean, I eventually resorted to suppressions and then gave up.
For example, `valgrind pango-view -t "This is a test"` finds 2 definitely lost blcoks, 304 indirectly lost and 21 possibly lost. The biggest leak has this backtrace:
```
==8098== 24,528 bytes in 3 blocks are still reachable in loss record 948 of 948
==8098==    at 0x4C2FB6B: malloc (vg_replace_malloc.c:299)
==8098==    by 0x5C29924: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C43FCD: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x7E14037: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E14E0B: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E12A52: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E13714: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E172DC: XML_ParseBuffer (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x5C42B52: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C42F85: FcConfigParseAndLoad (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C42FF7: FcConfigParseAndLoad (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C4316D: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x7E14037: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E14E0B: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E12A52: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E13714: ??? (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x7E172DC: XML_ParseBuffer (in /usr/lib64/libexpat.so.1.6.7)
==8098==    by 0x5C42B52: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C42F85: FcConfigParseAndLoad (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C35D53: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C35FA5: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C29A56: ??? (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x5C2BAE4: FcConfigSubstituteWithPat (in /usr/lib64/libfontconfig.so.1.10.1)
==8098==    by 0x54B3C27: ??? (in /usr/lib64/libpangocairo-1.0.so.0.4000.13)
==8098==    by 0x5094761: ??? (in /usr/lib64/libpangoft2-1.0.so.0.4000.13)
==8098==    by 0x4E54416: ??? (in /usr/lib64/libpango-1.0.so.0.4000.13)
==8098==    by 0x4E55B9F: pango_itemize_with_base_dir (in /usr/lib64/libpango-1.0.so.0.4000.13)
==8098==    by 0x4E5E55A: ??? (in /usr/lib64/libpango-1.0.so.0.4000.13)
==8098==    by 0x4E6042B: ??? (in /usr/lib64/libpango-1.0.so.0.4000.13)
==8098==    by 0x4E60900: pango_layout_get_pixel_extents (in /usr/lib64/libpango-1.0.so.0.4000.13)
==8098==    by 0x10D609: ??? (in /usr/bin/pango-view)
==8098==    by 0x10E1C7: ??? (in /usr/bin/pango-view)
==8098==    by 0x110965: ??? (in /usr/bin/pango-view)
==8098==    by 0x10CDDE: ??? (in /usr/bin/pango-view)
==8098==    by 0x66F7009: (below main) (in /usr/lib64/libc-2.26.so)
```
The function `FcConfigParseAndLoad` sounds a lot like it is loading the fontconfig configuration in `/etc/fonts` (this is just a guess!). So I wonder why you aren't seeing this memory leak nor do you need to call `FcFini`.

Also, yes you can use `cairo_debug_reset_static_data` with your own caching, you just have to make sure to clear your own cache before. Actually, your patch does *introduce* a memory leak since your cache is never cleared. ;-)

Finally, feel free to ignore me. @stapelberg asked for some comments and I left them, but I am not the one to decide what goes into i3lock and what not.

stapelberg: I’m definitely on board with the approach to first try hard and fix the memory leaks in cairo :)

psychon:
Okay, one last thing:
```
i3lock: version 2.9.1 (2017-06-21) © 2010 Michael Stapelberg
```
I'm not sure how many leaks were fixes since this version, but from a quick scroll through the output of `valgrind --leak-check=full --show-reachable=yes i3lock`, most of the leaks are through libxkbcommon, `cairo_xcb_surface_create`, xcb internals and dynamic linker internals. However, I did not check everything since the report is way too long:
```
==8310== LEAK SUMMARY:
==8310==    definitely lost: 256 bytes in 1 blocks
==8310==    indirectly lost: 107 bytes in 4 blocks
==8310==      possibly lost: 9,652 bytes in 113 blocks
==8310==    still reachable: 741,020 bytes in 5,901 blocks
==8310==         suppressed: 0 bytes in 0 blocks
```
Is it just me, or are ASAN and valgrind reporting wildly different results?

unlock_indicator.c
PandorasFox: > Also, yes you can use cairo_debug_reset_static_data with your own caching, you just have to make sure to clear your own cache before

ah, yeah, but if we've got to clear our cache every tick, it's not much of a cache, then :P

> Actually, your patch does introduce a memory leak since your cache is never cleared. ;-)

Yeah, I know. There's no real elegant way of doing this all without either caching nothing (a bit less than ideal) or adding some at-exit cleanup to i3lock (probably way more complicated than needed)

stapelberg: We’d prefer to not have exit cleanups unless absolutely necessary (e.g. to aid leak-checker tools). The OS cleans up when a process exits, that’s good enough.

PandorasFox:
Yeah, I get wildly different stuff with ASAN and valgrind, but I think that's due to an ASAN-compiled program running with non-ASAN libs. I was doing some debugging for another project and ended up compiling fontconfig with ASAN and ended up getting much more similar outputs after then.

I'm guessing it's just due to how ASAN gets built into the binary.

unlock_indicator.c
psychon: > The OS cleans up when a process exits, that’s good enough.

Which means that a memory leak is only a bug if it increases over time, not if it is a static cache like it is (or at least seems to be) here?

stapelberg: I’d say so, yes. (Again: with the caveat that if a static leak is not recognized as such by a memory leak checker, it might still be good to eliminate it to get a more useful report.)

PandorasFox: Yeah, I was pretty much making this pull with the goal of having a clean baseline going forwards.

psychon:
[This patch](https://gist.github.com/psychon/5a76f6a97430736c1a77bcc32bbf0949) makes valgrind go from
```
==15126== LEAK SUMMARY:
==15126==    definitely lost: 256 bytes in 1 blocks
==15126==    indirectly lost: 107 bytes in 4 blocks
==15126==      possibly lost: 9,652 bytes in 113 blocks
==15126==    still reachable: 740,708 bytes in 5,896 blocks
==15126==         suppressed: 0 bytes in 0 blocks
```
to
```
==17259== LEAK SUMMARY:
==17259==    definitely lost: 144 bytes in 3 blocks
==17259==    indirectly lost: 0 bytes in 0 blocks
==17259==      possibly lost: 0 bytes in 0 blocks
==17259==    still reachable: 272,496 bytes in 6 blocks
==17259==         suppressed: 0 bytes in 0 blocks
```
Remaining leaks are:
- An actual bug in cairo (`cairo_debug_reset_static_data` misses one cache which has 262,176 bytes and thus accounts for most of the remaining memory; thanks for making me spot this).
- Weird stuff called from `_dl_init` which I will just ignore.
- One timer that is still pending and cannot easily be `free`d (see patch; this is also where the `definite lost` comes from).

I'm not quite sure what my point here is, but it should be something like "there is no actual memory leak in cairo". I don't know why ASAN manages to find some leaks, but ignores all the other ones that valgrind finds, but... well, perhaps it is a good idea to figure out why ASAN only complains about some particular leak before adding 100 lines of code to make ASAN stop complaining.

stapelberg:
@PandorasFox Sorry for letting this PR linger for so long. Could you resolve the conflicts please? We can go ahead with merging it then.

PandorasFox:
I'll look into that later. I've been meaning to dig into the fontconfig errors some more and figure out if there's really a consistent way to resolve its errors (or if there's been a subsequent fontconfig release that's fixed them).

I noticed [this commit](https://github.com/freedesktop/fontconfig/commit/9f1b92f27f9259aa69e5387656cb7d4c1b305a98) and am unsure if the rewrite/extra complexity is really needed anymore.

PandorasFox:
I'm currently on a long flight and my laptop died shortly after I pushed that rebase. I'll do more testing on this later this weekend.

PandorasFox:
I built fontconfig with `-g -fsanitize=address -fno-omit-frame-pointer` from [this commit](https://github.com/behdad/fontconfig/commit/5b41ded2b0ddb98a07ac86264b94403cb7a0fd82). When I ran the current `i3lock` master pulling in my local fontconfig build, I got no errors:

`LD_PRELOAD=/usr/lib/libasan.so:/usr/local/lib/libfontconfig.so i3lock --nofork` -> no output. Running `i3lock --nofork` (using Arch's fontconfig 2.13.1+12+g5f5ec56-1) produces errors.

When running my fork vs the current master against Arch's fontconfig, the current master produces some ASAN leak warnings in the older fontconfig:

```
=================================================================
==22665==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x7f7b94b2f019 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:86
    #1 0x7f7b92e40345  (/usr/lib/libfontconfig.so.1+0x21345)

Indirect leak of 96 byte(s) in 3 object(s) allocated from:
    #0 0x7f7b94b2f231 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:95
    #1 0x7f7b92e40989  (/usr/lib/libfontconfig.so.1+0x21989)

Indirect leak of 11 byte(s) in 1 object(s) allocated from:
    #0 0x7f7b94a77f01 in __interceptor_strdup /build/gcc/src/gcc/libsanitizer/asan/asan_interceptors.cc:405
    #1 0x7f7b92e3fc25 in FcValueSave (/usr/lib/libfontconfig.so.1+0x20c25)

SUMMARY: AddressSanitizer: 363 byte(s) leaked in 5 allocation(s).
```

So - it definitely seems that the overall issue was with fontconfig and this pull only really resolves the symptoms of it. I don't really feel that this should be necessary for correctness on i3lock's part, but I guess it might still be nice to have so that font handling is broken out and a bit easier to audit (maybe?) instead of relying on cairo functions. I'm personally leaning towards closing this.

stapelberg:
> So - it definitely seems that the overall issue was with fontconfig and this pull only really resolves the symptoms of it. I don't really feel that this should be necessary for correctness on i3lock's part

Let’s close this, then. The less code we have, the better :).

Thanks for looking into this!

PR #171, opened 2018-01-12
Check PNG file compliance

### LINKED ISSUE #114: Providing an invalid PNG gives confusing error message "out of memory"

eplanet:
This fixes #114 with a quick sanity check on supplied PNG file:
- File existence
- Header is PNG

Tested successfully with:
- Nonexistent file
- Empty file
- File containing less than 8 bytes (header size)
- File containing more than 8 bytes but not PNG
- PNG file

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -19,6 +19,7 @@
 #include <xcb/xcb.h>
 #include <xcb/xkb.h>
 #include <err.h>
+#include <errno.h>
 #include <assert.h>
 #ifdef __OpenBSD__
 #include <bsd_auth.h>
@@ -632,6 +633,36 @@ void handle_screen_resize(void) {
     redraw_screen();
 }
 
+static bool verify_png_image(const char *image_path) {
+    if (!image_path) {
+        return false;
+    }
+
+    /* Check file exists and has correct PNG header */
+    FILE *png_file = fopen(image_path, "r");
+    if (png_file == NULL) {
+        fprintf(stderr, "Image file path \"%s\" cannot be opened: %s\n", image_path, strerror(errno));
+        return false;
+    }
+    unsigned char png_header[8];
+    memset(png_header, '\0', sizeof(png_header));
+    int bytes_read = fread(png_header, 1, sizeof(png_header), png_file);
+    fclose(png_file);
+    if (bytes_read != sizeof(png_header)) {
+        fprintf(stderr, "Could not read PNG header from \"%s\"\n", image_path);
+        return false;
+    }
+
+    // Check PNG header according to the specification, available at:
+    // https://www.w3.org/TR/2003/REC-PNG-20031110/#5PNG-file-signature
+    static unsigned char PNG_REFERENCE_HEADER[8] = { 137, 80, 78, 71, 13, 10, 26, 10 };
+    if (memcmp(PNG_REFERENCE_HEADER, png_header, sizeof(png_header)) != 0) {
+        fprintf(stderr, "File \"%s\" does not start with a PNG header. i3lock currently only supports loading PNG files.\n", image_path);
+        return false;
+    }
+    return true;
+}
+
 #ifndef __OpenBSD__
 /*
  * Callback function for PAM. We only react on password request callbacks.
@@ -1012,7 +1043,7 @@ int main(int argc, char *argv[]) {
     xcb_change_window_attributes(conn, screen->root, XCB_CW_EVENT_MASK,
                                  (uint32_t[]){XCB_EVENT_MASK_STRUCTURE_NOTIFY});
 
-    if (image_path) {
+    if (verify_png_image(image_path)) {
         /* Create a pixmap to render on, fill it with the background color */
         img = cairo_image_surface_create_from_png(image_path);
         /* In case loading failed, we just pretend no -i was specified. */
@@ -1021,8 +1052,8 @@ int main(int argc, char *argv[]) {
                     image_path, cairo_status_to_string(cairo_surface_status(img)));
             img = NULL;
         }
-        free(image_path);
     }
+    free(image_path);
 
     /* Pixmap on which the image is rendered to (if any) */
     xcb_pixmap_t bg_pixmap = draw_image(last_resolution);


eplanet:
Anyone for review? That should be pretty quick.

eplanet:
@stapelberg @Airblader ? Thanks!

i3lock.c
```diff
@@ -1009,13 +1009,29 @@ int main(int argc, char *argv[]) {
...
     if (image_path) {
-        /* Create a pixmap to render on, fill it with the background color */
-        img = cairo_image_surface_create_from_png(image_path);
-        /* In case loading failed, we just pretend no -i was specified. */
-        if (cairo_surface_status(img) != CAIRO_STATUS_SUCCESS) {
-            fprintf(stderr, "Could not load image \"%s\": %s\n",
-                    image_path, cairo_status_to_string(cairo_surface_status(img)));
-            img = NULL;
+        /* Check file exists and has correct PNG header */
+        if (access(image_path, F_OK) == -1) {
```
stapelberg: Remove this conditional and check `fopen`’s error instead please

i3lock.c
```diff
@@ -1009,13 +1009,29 @@ int main(int argc, char *argv[]) {
...
+            fprintf(stderr, "Image file path \"%s\" doesn't exist\n", image_path);
+        } else {
+            FILE* png_file = fopen(image_path, "r");
```
stapelberg: Please refactor this code into a helper function.

i3lock.c
```diff
@@ -1009,13 +1009,29 @@ int main(int argc, char *argv[]) {
...
+            unsigned char png_header[8];
+            memset(png_header, 0, 8);
+            fread(png_header, 1, 8, png_file);
```
stapelberg: Check `fread`’s errors please.

i3lock.c
stapelberg: Nit: `FILE *png_file` for consistency (star at the variable name, not the type)

i3lock.c
```diff
@@ -1009,13 +1009,29 @@ int main(int argc, char *argv[]) {
...
+            fclose(png_file);
+
+            if (png_header[0] != 137 || png_header[1] != 80 || png_header[2] != 78 || png_header[3] != 71
```
stapelberg: nit: one check per line please:
```c
if (png_header[0] != 137 ||
    png_header[1] != 80 ||
…
```

Also, can you add a comment describing where you got these values from?

i3lock.c
```diff
@@ -1009,13 +1009,29 @@ int main(int argc, char *argv[]) {
...
+                        || png_header[4] != 13 || png_header[5] != 10 || png_header[6] != 26 || png_header[7] != 10) {
+                fprintf(stderr, "Provided PNG image \"%s\" is incorrect\n", image_path);
```
stapelberg: Let’s make this error message more descriptive. How about:

File "%s" does not start with a PNG header. i3lock currently only supports loading PNG files.

i3lock.c
psychon: Honest question: Would it make sense to have an array with the expected values and just `memcmp` this with the actual values?

stapelberg:
Yeah, using an array sounds good to me. Also, please refactor the new logic into a helper function (which should clean up the control flow).

eplanet:
I updated the PR accordingly. I hope I got well what you expected as a helper function. Feel free to ask for more modifications.

Thanks!

eplanet:
Everything's set! Looks very clean indeed. Hope it's cool this time :-).

i3lock.c
```diff
@@ -632,6 +632,36 @@ void handle_screen_resize(void) {
     redraw_screen();
 }
 
+static bool verify_png_image(char *image_path) {
```
stapelberg: Let’s make this `const char *image_path`, as it’s not being modified

i3lock.c
```diff
@@ -632,6 +632,36 @@ void handle_screen_resize(void) {
...
+    if (!image_path) {
+        return false;
+    }
+
+    /* Check file exists and has correct PNG header */
+    FILE *png_file = fopen(image_path, "r");
+    if (png_file == NULL) {
+        fprintf(stderr, "Image file path \"%s\" cannot be opened\n", image_path);
```
stapelberg: Let’s tell the user why the file could not be opened: please include strerror(errno) in the message. Please also verify that the error message is printed as you’d expect when specifying a file that doesn’t access (or where you don’t have read permission).

eplanet: Just added that. For the 2 tests you mention, here are the messages:
- Image file path "foo" cannot be opened: "No such file or directory"
- Image file path "bar" cannot be opened: "Permission denied"

i3lock.c
```diff
@@ -632,6 +633,36 @@ void handle_screen_resize(void) {
...
+    if (png_file == NULL) {
+        fprintf(stderr, "Image file path \"%s\" cannot be opened: \"%s\"\n", image_path, strerror(errno));
+        return false;
+    }
+    unsigned char png_header[8];
+    memset(png_header, '\0', sizeof(png_header));
+    int bytes_read = fread(png_header, 1, sizeof(png_header), png_file);
+    fclose(png_file);
+    if (bytes_read != sizeof(png_header)) {
+        fprintf(stderr, "Could not read the size of PNG header in file \"%s\"\n", image_path);
```
stapelberg: Sorry, missed this spot in the previous review, but let’s also make sure that this error message is as good as it can be. I suggest:

fprintf(stderr, "Could not read PNG header from \"%s\": %s\n", image_path, strerror(errno));

i3lock.c
```diff
@@ -632,6 +633,36 @@ void handle_screen_resize(void) {
...
 
+static bool verify_png_image(const char *image_path) {
+    if (!image_path) {
+        return false;
+    }
+
+    /* Check file exists and has correct PNG header */
+    FILE *png_file = fopen(image_path, "r");
+    if (png_file == NULL) {
+        fprintf(stderr, "Image file path \"%s\" cannot be opened: \"%s\"\n", image_path, strerror(errno));
```
stapelberg: No need for the quotes around the error message. The quotes are nice to ensure the file name is clearly separated from the remainder of the message, but strerror’s return value is supposed to blend nicely into error messages. Let’s remove the quotes.

i3lock.c
eplanet: Actually `fread` doesn't set `errno` to a valid value, I have "`No such file or directory`" which doesn't make sense. I propose to stick to:
```cpp
fprintf(stderr, "Could not read PNG header from \"%s\"\n", image_path);
```

eplanet:
Great! :-)

PR #177, opened 2018-02-11
return value corrected

SegFault42:
xcb_connect function never return NULL (see the man here : https://xcb.freedesktop.org/manual/group__XCB__Core__API.html#ga094470586356d1764e69c9a1882966c3)

"Callers need to use xcb_connection_has_error() to check for failure."

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -807,8 +807,7 @@ static void raise_loop(xcb_window_t window) {
     xcb_generic_event_t *event;
     int screens;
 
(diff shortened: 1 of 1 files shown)


Airblader:
I was wondering at first whether this had changed over time since I don't recall ever seeing a (serious) X11 client using XCB that doesn't check this; however, at least awesome also doesn't check it, so combined with the documentation I assume this should be OK.

Issue #175, opened 2018-01-29
getpwuid() fails


jbeigel:
I try to use i3lock on a Linux system with LDAP user account. Later in the code, this seems to be handled correctly via PAM, but right at the start after the option parsing, getpwuid() is called, fails and i3lock immediately exit with the (somewhat funny) error "i3lock: getpwuid() failed: Success"

SegFault42:
It's because getpwuid return NULL and errno is set to 0 (Success).

If you read the manual, you can see that (in RETURN VALUE and ERRORS section):

```
If  an error occurs, errno is set appropriately.

ERRORS
       0 or ENOENT or ESRCH or EBADF or EPERM or ...
              The given name or uid was not found.
```

I don't know how LDAP user acount work with getpwuid().

Z1NZ:
Hi @jbeigel @SegFault42  

We can have  more information about issue like linux (32/64) ?  distribution of linux ?
Did you compile source code ?

I find similar issue on stackoverflow 
https://stackoverflow.com/questions/18570177/getpwuid-returns-null-for-ldap-user

stapelberg:
We could totally add a fallback code path which used the $USER environment variable (if set) when getpwuid fails. @jbeigel would you be able to test such a patch in your environment?

PR #178, opened 2018-02-14
 confusing error message getpwuid() 

SegFault42:
In issue #175 the function getpwuid() sets errno to zero while it returns null.
I think that in this case it is not necessary to display errno because it is confusing.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -807,8 +807,7 @@ static void raise_loop(xcb_window_t window) {
     xcb_generic_event_t *event;
     int screens;
 
-    if ((conn = xcb_connect(NULL, &screens)) == NULL ||
-        xcb_connection_has_error(conn))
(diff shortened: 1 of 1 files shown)


i3lock.c
```diff
@@ -879,8 +879,10 @@ int main(int argc, char *argv[]) {
         {"show-failed-attempts", no_argument, NULL, 'f'},
         {NULL, no_argument, NULL, 0}};
 
-    if ((pw = getpwuid(getuid())) == NULL)
+    if ((pw = getpwuid(getuid())) == NULL && errno != 0)
         err(EXIT_FAILURE, "getpwuid() failed");
+    else
```
Airblader: I don't really understand this change – this will now just always error.

Airblader:
Since this PR is quite old (sorry for the lack of feedback), small and seems like an odd change I will close for now. If you'd still like to make this change, feel free to comment on my concern above and we can reopen.

PR #179, opened 2018-03-06
Fix memory leak when grabbing fails

karulont:
Noticed this issue when running with valgrind.

When we retry to grab, we always create new reply objects without freeing them.
This PR fixes that.

### DIFF
--- a/xcb.c
+++ b/xcb.c
@@ -203,6 +203,9 @@ bool grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
             break;
         }
 
+        /* In case the grab failed, we still need to free the reply */
+        free(preply);
+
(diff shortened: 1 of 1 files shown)


xcb.c
```diff
@@ -203,6 +203,11 @@ bool grab_pointer_and_keyboard(xcb_connection_t *conn, xcb_screen_t *screen, xcb
             break;
         }
 
+        /* In case the grab failed, we still need to free the reply */
+        if (preply) {
```
orestisf1993: You don't need to check for NULL with free

orestisf1993:
Up.

You need to leave a comment or else they don't get a notification.

Airblader:
Indeed. Thanks to both of you!

PR #183, opened 2018-05-06
Add command line option to lock TTY switching

LorianColtof:


i3lock.c
```diff
@@ -942,9 +953,16 @@ int main(int argc, char *argv[]) {
             case 'f':
                 show_failed_attempts = true;
                 break;
+            case 'l':
+#if defined(__linux__)
+                lock_tty_switching = 1;
```
Airblader: nit: `true`

i3lock.c
```diff
@@ -859,6 +864,11 @@ int main(int argc, char *argv[]) {
     int ret;
     struct pam_conv conv = {conv_callback, NULL};
 #endif
+#if defined(__linux__)
+    int lock_tty_switching = 0;
```
Airblader: nit: This should be a `bool`

PR #185, opened 2018-05-11
Respect Xft.dpi for determining the unlock indicator’s scale factor

### LINKED ISSUE #184: i3lock should respect Xft.dpi
Currently, i3lock calculates the DPI by querying the display height (in mm) and resolution (in pixels). This is what we used to do in i3, but we should respect Xft.dpi if set, and only fall back to this mechanism.

stapelberg:
fixes #184

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 # Use Ubuntu 14.04 (trusty), as per http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/
-sudo: required
+sudo: false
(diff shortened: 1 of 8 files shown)


stapelberg:
We don’t use docker in this repository’s .travis.yml yet, so I’ll need to change that before we can get libxcb-xrm*

Airblader:
LGTM, waiting for the Travis changes :)

stapelberg:
Travis changes done. Do you want to have another look at this PR?

PR #188, opened 2018-05-20
Added ISSUE_TEMPLATE.md

Airblader:
This adds an issue template for i3lock. The content of the template is inspired by angular/angular. I'd like to update the template in i3 in a similar fashion, perhaps we could also update the bot to use the checkboxes to deal with tags more easily (it's currently not transparent to users how to open issues which should be tagged as feature requests) and close questions that are missing version information.


Airblader:
(I updated this similarly to the discussions in the main repository)

PR #190, opened 2018-06-18
Execute a command when a maximum of auth-attempts is exceeded

VipeOut23:
This branch added the commandline options "-x" and "-m".
With "-x" a command is specified, which will be executed, if the maximum ammount of authorization attempts is exceeded. The option "-m" is used to set the maximum allowed retries.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -106,6 +106,16 @@ your computer with the enter key.
 Show the number of failed attempts, if any.
 
 .TP
+.BI \-x\  command \fR,\ \fB\-\-max\-failed\-attempts\-exec= command 
+Run a command, when the the maximum ammount of failed authentication
+attempts (default 3; can be specified with \-m) is exceeded.
+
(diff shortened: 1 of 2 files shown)


Airblader:
Hi, thanks for the PR. We generally recommend that you first open an issue so we can discuss potential features before implementing them in case we'd like to not add it to i3lock.

Since it's too late for that, let's discuss it here. What usecase did you have in mind for the command you added? Also, wouldn't it be better to handle monitoring of the authentication within the software that handles that, namely pam?

VipeOut23:
Hi, thanks for your response. I had in mind, that it would be useful on portable devices, so the device may shut down (in case of an encrypted disk) or somehow log unwanted access attempts. I didn't mind pam, that would be a much better way to achieve this. So it's probably not the right way to let i3lock handle this.
(Sorry for directly openign a PR, I'm not really active on github)

PR #192, opened 2018-06-25
Keep only the last attempt for retry_verification

### LINKED ISSUE #191: Scrambled passwords during STATE_AUTH_WRONG
Hi,

Typing too fast has a strange undefined behavior in the AUTH_WRONG state.
This happens often to me as I type my password fairly quickly but, for some reason, I tend to make many mistakes.

When in AUTH_WRONG state, hitting enter (or any validation key) sets `retry_verification = true;` https://github.com/i3/i3lock/blob/ad680b40192f665dd7af4d62ab6996318e275cb6/i3lock.c#L435
If more characters are typed after that, they are added after the validated password, making it useless when verification is retried.

For example, with a password set to abcd. Typing `"xxxx\nabcd\nab"` within two seconds will
1. validate "xxxx" as a password attempt and fail. i3lock enters the AUTH_WRONG state and further attempts are locked for two seconds
2. "abcd" will be added to the password buffer
3. enter will set `retry_verification` to `true`.
4. "ab" will be added to the password buffer
5. after the two seconds delay, "abcdab" will be considered the next password attempt, and fail.

This can go on for some time, as the logical continuation is to type "cd\n", which will obviously spoil the next attempt, and so on.

I would suggest to either
1. keep the last _submitted_ password for `retry_verification` and keep extra characters in a separate buffer.
2. discard the previously submitted password if extra characters are entered after submitting a password (see #192).

layus:
Closes #191

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -445,6 +445,12 @@ static void handle_key_press(xcb_key_press_event_t *event) {
             return;
         default:
             skip_repeated_empty_password = false;
+            // A new password is being entered, but a previous one is pending.
+            // Discard the old one and clear the retry_verification flag.
+            if (retry_verification) {
+                retry_verification = false;
+                clear_input();
+            }
     }
 
     switch (ksym) {


layus:
@Airblader Would you mind reviewing this ? Then change is very straightforward.

i3lock.c
```diff
@@ -445,6 +445,12 @@ static void handle_key_press(xcb_key_press_event_t *event) {
             return;
         default:
             skip_repeated_empty_password = false;
+            // A new password i being entered, but a previous one is pending.
```
Airblader: Typo: »i« → »is«

layus:
Oops, sorry. Updated!

PR #197, opened 2018-08-23
Add command line options for hiding text and changing the color of the unlock indicator

arch1t3cht30:
- Added a command line option to hide the status messages like 'verifying...' and 'wrong!'
- Added a command line option to change the color of the unlock indicator ring, the default being the previous color. The color of the arcs highlighted for key presses is calculated automatically as a desaturated and lighter version of the given ring color, in a way that agrees with the original hardcoded highlight color.


Airblader:
Hi, thanks for sending the PR. We're not really looking to extend i3lock's visual configuration at the moment. Because of that I'll have to close the PR, sorry. Thanks for understanding!

PR #213, opened 2018-12-17
Rename Win Key to Super Key

trickeydan:
In most Linux environments, this key is referred to as "Super" to prevent infringing Microsoft's trademarks. Whilst "Win" is not a trademark of Microsoft it would be nice to be consistent.

https://en.wikipedia.org/wiki/Super_key_(keyboard_button)


orestisf1993:
we could do the same in i3

PR #214, opened 2018-12-31
Exit with a failure if exit() is not expected by us

sandsmark:
This allows wrapping i3lock in a shell script that handles crashes or
unexpected exits, e. g. from a library like in CVE-2016-2312.

Same idea as https://github.com/i3/i3lock/pull/108 by allowing people to wrap i3lock in a shell script to handle crashes/unexpected exits.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -102,6 +102,19 @@ bool tile = false;
 bool ignore_empty_password = false;
 bool skip_repeated_empty_password = false;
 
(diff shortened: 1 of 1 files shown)


Airblader:
Instead of using such a flag, would this be caught by instead setting up sigterm/sigkill handlers?

Airblader:
Closing due to inactivity

Issue #204, opened 2018-11-10
windows get shown on top when compton is active


bebehei:
## Current Behavior

Windows appearing after i3lock's start and in combination with `compton` will get shown on top. When `compton` is not active, the behavior is not seen.

Here's a screenshot of a clean Ubuntu VM with the reproduction instructions executed:

![screenshot-bionic](https://user-images.githubusercontent.com/2073458/48302285-2eecdf80-e4fb-11e8-8eec-b85315be72e3.png)

## Expected Behavior

Windows shall get drawn below i3lock.

## Reproduction Instructions

```
# On a clean Ubuntu 18.04 installation
$> sudo apt install i3-wm i3lock i3status dunst compton

# And after i3 login:
$> compton -c &
$> dunst &
$> (sleep 7 && notify-send a b) &
$> i3lock
```

## Environment
Output of `i3lock --version`:
```
$> i3lock --version
i3lock: version 2.10 (2017-11-25) © 2010 Michael Stapelberg
```

# Related

#130 

Originally reported on dunst-project/dunst#553

Airblader:
@stapelberg Why did we want to open a jew issue here? We fixed it, it broke stuff, we reverted it. I'm not sure what else there is to do?

stapelberg:
Can’t we find a new fix which doesn’t break? Or at least document the defect.

bebehei:
Oh, shit, I forgot the most important part: I cannot reproduce it on ArchLinux.

I've also cloned now i3lock's repo and tested the following tags/branches: `2.9.1`, `2.10`, `2.11.1`, `master`. All show the same bad behavior in the Ubuntu VM.

I believe it's a problem in the coupling of some library, which is used by Ubuntu 18.04.

alskjdjutcfrnju

bebehei:
Sorry, my browser lagged. About "the library": I can definitely exclude compton: ArchLinux uses a fork of compton. I've compiled the fork on Ubuntu 18.04, but it doesn't help either.

Airblader:
@stapelberg The author of xsecurelock commented on #96 and said that they also have the problem with different compositors and different configs thereof. His idea (for xsecurelock) was to maybe make it a flag whether to use the overlay window or not. That's about the only thing I can think of.

4goettma:
I can confirm the descriped behaviour on Arch Linux.
```
$> i3lock --version
i3lock: version 2.11.1 © 2010 Michael Stapelberg
```
```
$> compton --version
[...]

PR #215, opened 2018-12-31
Fix covering of composited notifications (#204)

sandsmark:
Just cherry-picked in my old fix, but this time hidden behind an option like suggested in the bug report.

Use the XComposite extension to get the composite overlay window,
instead of just using the normal root window. This ensures that
composited windows are covered.

This time, we hide it behind a config option because it might break
things more than it fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Requirements
 - libxcb-util
 - libpam-dev
 - libcairo-dev
+- libxcb-composite0
 - libxcb-xinerama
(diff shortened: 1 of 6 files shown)


Airblader:
We didn't really have a decision on how to proceed with this yet, and I brought it up in #204 again. I'll close this PR for now until we have a decision here – if we do opt for the flag, we can just reopen this PR. :-)

PR #216, opened 2019-01-01
Remove \n from errx messages

pclouds:
errx() already appends \n internally. "\n" in the error message will result in a blank line after the message.


pclouds:
> This can be done for https://github.com/i3/i3 as well

Would you like me to open a pull request to fix that as well, or you're fixing it already? I'm OK either way.

orestisf1993:
Open a PR if you'd like, I didn't fix it yet.

PR #218, opened 2019-01-14
added wm_class property

ivanvig:
Right now i3lock has only a wm_name property but lacks of wm_class which is useful to, for example, set compton rules for i3lock's window.

### DIFF
--- a/xcb.c
+++ b/xcb.c
@@ -149,6 +149,15 @@ xcb_window_t open_fullscreen_window(xcb_connection_t *conn, xcb_screen_t *scr, c
                         strlen(name),
                         name);
 
+    xcb_change_property(conn,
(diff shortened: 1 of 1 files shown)


Airblader:
Thanks for the PR! I'm definitely in favor of adding this, but I think we could do entirely without the new macros here as they just add a level of redirection when reading the code that we don't need. Could you remove them and just use string literals instead? Be carefull for the length of `"i3lock\0i3lock\0"` to use `2 * (strlen("i3lock") + 1)`.

Also, please squash the commits into a single commit (and force-push the branch) and drop a comment here so that we get notified. Thank you!

ivanvig:
I think is done, went back to master and only added the code block corresponding to the property

Airblader:
Thank you!

PR #220, opened 2019-01-31
Added hook when login failed/succeed.

TheKinrar:
*This is a PR by https://mastodon.xyz/@Darks who does not have a GitHub account.*

Exemples:
* Take a picture with webcam when login failed
* Send a notification if someone tried to login

Environnement variable I3L_ATTEMPTS can be used in the hook scripts.
Ex: `notifify-send -u critical "$I3L_ATTEMPTS failed login attempts"`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+2018-10-18 i3lock 2.12
+
+ • Add hooks when login failed/succeed
+
 2018-10-18 i3lock 2.11.1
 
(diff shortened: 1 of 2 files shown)


Airblader:
Thanks for the PR; however, we recommend you discuss feature ideas in issues first to see whether a patch would be accepted. In this instance, this has been discussed in the past, see #61 and #190, so I'm closing for this reason.

If there's new/different arguments to be made, please raise an issue to discuss it. Thanks!

PR #221, opened 2019-02-06
Use XKB_KEY_F35 as image reload.

Yoshqu:
For fancier dynamic backgrouds, this simple patch adds image reload for F35 key - which can be easily sent via xdotool. Had to move png loaders up.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -97,6 +97,7 @@ static uint8_t xkb_base_event;
 static uint8_t xkb_base_error;
 static int randr_base = -1;
 
+static char *image_path = NULL;
 cairo_surface_t *img = NULL;
 bool tile = false;
 bool ignore_empty_password = false;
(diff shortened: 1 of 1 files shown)


Airblader:
Hi,

thanks for your PR. However, i3lock is not intended to be about fanciness, and we try to keep the surface in this direction as small as possible to avoid introducing unnecessary risks. Therefore I'm going to close this PR. We generally recommend that you first raise an issue to discuss new features to see if we'd accept it or how it could look like. Thanks for understanding!

stapelberg:
To add to @Airblader’s message:

Thanks for your contribution!

However, this change seems too much like a security risk to me. One of i3lock’s design goals is to do as little as possible while the screen is locked. Reloading images might circumvent the screen lock due to poor error handling (not necessarily in your code, but in the entire stack) or denial of service issues (out of memory due to an otherwise innocuous memory leak), etc.

Thanks for your understanding,

Yoshqu:
No harm intended. I hope it has brought some smile :). This is beauty of i3 - i can expand as i see it (and it was too small for launching a discussion)

@stapelberg yes, it would require a lot more work to load/show image in forked process to make this secure - so it would be no go anyway....

BTW Thanks for your work.

PR #224, opened 2019-03-21
unlock indicator optimization

hissssst:
Small unlock indicator optimization was made. Now indicator updates only the area in the center of the screen, as it was requested in the comment in the code


unlock_indicator.c
```diff
@@ -6,26 +6,40 @@
...
 #include "randr.h"
-#include "dpi.h"
+#include "unlock_indicator.h"
+#include "xcb.h"
 
 #define BUTTON_RADIUS 90
 #define BUTTON_SPACE (BUTTON_RADIUS + 5)
 #define BUTTON_CENTER (BUTTON_RADIUS + 5)
 #define BUTTON_DIAMETER (2 * BUTTON_SPACE)
+/* Color constants and stuff */
```
Airblader: Let's remove this comment and maybe instead ive the constants more descriptive names as to what they represent.

unlock_indicator.c
```diff
@@ -86,10 +100,9 @@ auth_state_t auth_state;
  * resolution and returns it.
  *
  */
-xcb_pixmap_t draw_image(uint32_t *resolution) {
+
+xcb_pixmap_t draw_image_with_sf_and_diam(uint32_t *resolution, const double scaling_factor, int button_diameter_physical) {
```
Airblader: I don't quite understand why this function was split like this since you call it with exactly the same sf and button diameter in both places?

unlock_indicator.c
```diff
@@ -332,17 +347,27 @@ xcb_pixmap_t draw_image(uint32_t *resolution) {
...
  *
  */
 void redraw_screen(void) {
     DEBUG("redraw_screen(unlock_state = %d, auth_state = %d)\n", unlock_state, auth_state);
-    xcb_pixmap_t bg_pixmap = draw_image(last_resolution);
+    const double scaling_factor = get_dpi_value() / 96.0;
+    int button_diameter_physical = ceil(scaling_factor * BUTTON_DIAMETER);
+    xcb_pixmap_t bg_pixmap = draw_image_with_sf_and_diam(last_resolution, scaling_factor, button_diameter_physical);
     xcb_change_window_attributes(conn, win, XCB_CW_BACK_PIXMAP, (uint32_t[1]){bg_pixmap});
     /* XXX: Possible optimization: Only update the area in the middle of the
```
Airblader: This comment can now be removed, right?

unlock_indicator.c
hissssst: In `redraw_screen()` I created `scaling_factor` and `button_diameter_physical` variables to calculate actual area of lock to clear (which is passed to `xcb_clear_area()`). `draw_image` function computes `scaling_factor` and `button_diameter_physical` too. So I created new function which takes `scaling_factor` and `button_diameter_physical` as arguments so program wouldn't compute them twice in `redraw_screen`.

It is a small bit of optimization, so I think it can be removed to keep the code more *easy to understand*

unlock_indicator.c
hissssst: Will do

unlock_indicator.c
Airblader: AFAIK the DPI is always constant, therefore is the scaling factor and thus so is the button diameter. So maybe we should just make those global values?

hissssst: > AFAIK the DPI is always constant, therefore is the scaling factor and thus so is the button diameter. So maybe we should just make those global values?

Yes, `dpi` is constant, `BUTTON_DIAMETER` is defined, but `button_diameter_physical` is neither defined nor constant. We can make them global values but it is too complex to save one computation. Even this split was too much for a few operations

Airblader: > but button_diameter_physical is neither defined nor constant

But it's just a simple arithmetic operation between two constants, so it's effectively constant. But maybe for now let's undo splitting the function and then we can at least move ahead with this PR's main focus. :-)

Airblader:
@hissssst Please also squash your changes into one (or more, but meaningful) commits. Thanks! :-)

McKean:
out of curiosity, will this hit master anytime soon? Just asking as I'm running dual 4k screens and the delays are quite significant. Now I'm not sure if I should clone and install vs. update via package manager. Thanks!

stapelberg:
> out of curiosity, will this hit master anytime soon? Just asking as I'm running dual 4k screens and the delays are quite significant. Now I'm not sure if I should clone and install vs. update via package manager. Thanks!

Are you already using commit https://github.com/i3/i3lock/commit/467e46c077a5b7e38aca28e8a43da9bd393713ff? That made things faster for me.

If yes, could you please build i3lock with this unlock indicator change and confirm that it helps?

McKean:
Will do! Need to head back where my second monitor is, but I'll report back.

McKean:
@stapelberg looks like I was still using i3-lock-next and have not properly setup my prime gpu offloading. After switching back to vanilla i3lock and configuring my gpus properly the lag is gone! Thanks for the quick response. And thanks for the amazing locker.

stapelberg:
@hissssst is this no longer relevant?

Did you notice if this optimization made any noticeable/measurable difference?

(Sorry that reviews are going slow currently.)

PR #226, opened 2019-04-11
Add an option to read an image as raw bytes rather than PNG

mortie:
There's a lot of tools, such as [i3lock-fancy-rapid](https://github.com/yvbbrjdr/i3lock-fancy-rapid), which work by creating an image, encoding that image as PNG, and running `i3lock -i <path>`. I tried using that i3lock-fancy-rapid on my machine, which has a 4k display, and found that just encoding the PNG image and writing it to a file took seconds; longer than the actual blur in some cases.

This patch makes it possible to just dump a buffer of raw pixel bytes into i3lock. I made my own [fork of i3lock-fancy-rapid](https://github.com/mortie/i3lock-fancy-rapid/blob/b398782186bf6b2f9db4e276860de27a58c7e5d7/i3lock-fancy-rapid.c#L145-L163), and found that with my preferred level of blurring, it's twice as fast; going from 0.68 seconds to 0.33 seconds. Without the blur, the difference is even bigger; from 0.65 seconds to 0.2 seconds.

I completely understand if this is out of scope for i3lock or something, I'll just keep maintaining my own branch, but I really think the performance gains are worth it. It also fits the "you should do all pre-processing in external tools" idea.

### DIFF
--- a/i3lock.1
+++ b/i3lock.1
@@ -20,6 +20,8 @@ i3lock \- improved screen locker
 .RB [\|\-b\|]
 .RB [\|\-i
 .IR image.png \|]
+.RB [\|\-r
+.IR format \|]
 .RB [\|\-c
 .IR color \|]
 .RB [\|\-t\|]
@@ -73,6 +75,16 @@ verified or whether it is wrong).
 .TP
 .BI \-i\  path \fR,\ \fB\-\-image= path
 Display the given PNG image instead of a blank screen.
+
+.TP
+.BI \-r\  format \fR,\ \fB\-\-raw= format
+Read the image given by \-\-image as a raw image instead of PNG. The argument is the image's format
+as <width>x<height>:<pixfmt>. The supported pixel formats are 'native' and 'rgb'.
+The "rgb" pixel format expects a pixel to be three bytes; red, green, and blue.
+The "native" pixel format expects a pixel as a 32-bit (4-byte) integer in
+the machine's native endianness, with the upper 8 bits unused. Red, green and blue are stored in
+the remaining bits, in that order.
+Example: \-\-raw=1920x1080:rgb
 
 .TP
 .BI \-c\  rrggbb \fR,\ \fB\-\-color= rrggbb
--- a/i3lock.c
+++ b/i3lock.c
@@ -645,6 +645,119 @@ void handle_screen_resize(void) {
     redraw_screen();
 }
 
+static ssize_t read_raw_image_native(uint32_t *dest, FILE *src, size_t width, size_t height, int pixstride) {
+    ssize_t count = 0;
+    for (size_t y = 0; y < height; y++) {
+        size_t n = fread(&dest[y * pixstride], 1, width * 4, src);
+        count += n;
+        if (n < (size_t)(width * 4))
+            break;
+    }
+
+    return count;
+}
+
+static ssize_t read_raw_image_rgb(uint32_t *dest, FILE *src, size_t width, size_t height, int pixstride) {
+    unsigned char *buf = malloc(width * 3);
+    if (buf == NULL)
+        return -1;
+
+    ssize_t count = 0;
+    for (size_t y = 0; y < height; y++) {
+        size_t n = fread(buf, 1, width * 3, src);
+        count += n;
+        if (n < (size_t)(width * 3))
+            break;
+
+        for (size_t x = 0; x < width; ++x) {
+            int idx = x * 3;
+            dest[y * pixstride + x] = 0 |
+                                      (buf[idx + 0] << 16) |
+                                      (buf[idx + 1] << 8) |
+                                      (buf[idx + 2]);
+        }
+    }
+
+    free(buf);
+    return count;
+}
+
+static cairo_surface_t *read_raw_image(const char *image_path, const char *image_raw_format) {
+    cairo_surface_t *img;
+
+#define RAW_PIXFMT_MAXLEN 6
+#define STRINGIFY1(x) #x
+#define STRINGIFY(x) STRINGIFY1(x)
+    /* Parse format as <width>x<height>:<pixfmt> */
+    char pixfmt[RAW_PIXFMT_MAXLEN + 1];
+    size_t w, h;
+    const char *fmt = "%zux%zu:%" STRINGIFY(RAW_PIXFMT_MAXLEN) "s";
+    if (sscanf(image_raw_format, fmt, &w, &h, pixfmt) != 3) {
+        fprintf(stderr, "Invalid image format: \"%s\"\n", image_raw_format);
+        return NULL;
+    }
+#undef RAW_PIXFMT_MAXLEN
+#undef STRINGIFY1
+#undef STRINGIFY
+
+    /* Create image surface */
+    img = cairo_image_surface_create(CAIRO_FORMAT_RGB24, w, h);
+    if (cairo_surface_status(img) != CAIRO_STATUS_SUCCESS) {
+        fprintf(stderr, "Could not create surface: %s\n",
+                cairo_status_to_string(cairo_surface_status(img)));
+        return NULL;
+    }
+    cairo_surface_flush(img);
+
+    /* Use uint32_t* because cairo uses native endianness */
+    uint32_t *data = (uint32_t *)cairo_image_surface_get_data(img);
+    const int pixstride = cairo_image_surface_get_stride(img) / 4;
+
+    FILE *f = fopen(image_path, "r");
+    if (f == NULL) {
+        fprintf(stderr, "Could not open image \"%s\": %s\n",
+                image_path, strerror(errno));
+        cairo_surface_destroy(img);
+        return NULL;
+    }
+
+    /* Read the image, respecting cairo's stride, according to the pixfmt */
+    ssize_t size, count;
+    if (strcmp(pixfmt, "native") == 0) {
+        /* If the pixfmt is 'native', just read each line directly into the buffer */
+        size = w * h * 4;
+        count = read_raw_image_native(data, f, w, h, pixstride);
+    } else if (strcmp(pixfmt, "rgb") == 0) {
+        /* If the pixfmt is 'rgb', we have to convert it to the native pixfmt */
+        size = w * h * 3;
+        count = read_raw_image_rgb(data, f, w, h, pixstride);
+    } else {
+        fprintf(stderr, "Unknown raw pixel pixfmt: %s\n", pixfmt);
+        fclose(f);
+        cairo_surface_destroy(img);
+        return NULL;
+    }
+    cairo_surface_mark_dirty(img);
+
+    if (count < size) {
+        if (count < 0 || ferror(f)) {
+            fprintf(stderr, "Failed to read image \"%s\": %s\n",
+                    image_path, strerror(errno));
+            fclose(f);
+            cairo_surface_destroy(img);
+            return NULL;
+        } else {
+            /* Print a warning if the file contains less data than expected,
+             * but don't abort. It's useful to see how the image looks even if it's wrong. */
+            fprintf(stderr, "Warning: expected to read %zi bytes from \"%s\", read %zi\n",
+                    size, image_path, count);
+        }
+    }
+
+    fclose(f);
+    return img;
+}
+
 static bool verify_png_image(const char *image_path) {
     if (!image_path) {
         return false;
@@ -867,6 +980,7 @@ int main(int argc, char *argv[]) {
     struct passwd *pw;
     char *username;
     char *image_path = NULL;
+    char *image_raw_format = NULL;
 #ifndef __OpenBSD__
     int ret;
     struct pam_conv conv = {conv_callback, NULL};
@@ -890,6 +1004,7 @@ int main(int argc, char *argv[]) {
         {"help", no_argument, NULL, 'h'},
         {"no-unlock-indicator", no_argument, NULL, 'u'},
         {"image", required_argument, NULL, 'i'},
+        {"raw", required_argument, NULL, 'r'},
         {"tiling", no_argument, NULL, 't'},
         {"ignore-empty-password", no_argument, NULL, 'e'},
         {"inactivity-timeout", required_argument, NULL, 'I'},
@@ -902,7 +1017,7 @@ int main(int argc, char *argv[]) {
     if ((username = pw->pw_name) == NULL)
         errx(EXIT_FAILURE, "pw->pw_name is NULL.");
 
-    char *optstring = "hvnbdc:p:ui:teI:fl";
+    char *optstring = "hvnbdc:p:ui:r:teI:fl";
     while ((o = getopt_long(argc, argv, optstring, longopts, &longoptind)) != -1) {
         switch (o) {
             case 'v':
@@ -934,6 +1049,9 @@ int main(int argc, char *argv[]) {
             }
             case 'u':
                 unlock_indicator = false;
+                break;
+            case 'r':
+                image_raw_format = strdup(optarg);
                 break;
             case 'i':
                 image_path = strdup(optarg);
@@ -969,7 +1087,7 @@ int main(int argc, char *argv[]) {
                 break;
             default:
                 errx(EXIT_FAILURE, "Syntax: i3lock [-v] [-n] [-b] [-d] [-c color] [-u] [-p win|default]"
-                                   " [-i image.png] [-t] [-e] [-I timeout] [-f] [-l]");
+                                   " [-i image.png] [-r format] [-t] [-e] [-I timeout] [-f] [-l]");
         }
     }
 
@@ -1069,7 +1187,11 @@ int main(int argc, char *argv[]) {
     xcb_change_window_attributes(conn, screen->root, XCB_CW_EVENT_MASK,
                                  (uint32_t[]){XCB_EVENT_MASK_STRUCTURE_NOTIFY});
 
-    if (verify_png_image(image_path)) {
+    if (image_raw_format != NULL && image_path != NULL) {
+        /* Read image. 'read_raw_image' returns NULL on error,
+         * so we don't have to handle errors here. */
+        img = read_raw_image(image_path, image_raw_format);
+    } else if (verify_png_image(image_path)) {
         /* Create a pixmap to render on, fill it with the background color */
         img = cairo_image_surface_create_from_png(image_path);
         /* In case loading failed, we just pretend no -i was specified. */
@@ -1079,7 +1201,9 @@ int main(int argc, char *argv[]) {
             img = NULL;
         }
     }
+
     free(image_path);
+    free(image_raw_format);
 
     /* Pixmap on which the image is rendered to (if any) */
     xcb_pixmap_t bg_pixmap = draw_image(last_resolution);


Airblader:
I somewhat agree that compared to other proposals of supporting JPEG, BMP, … this is at least a fairly basic/generic approach. At the same time I'm not sure how much this would be used as most people don't deal with raw data. @stapelberg What do you think?

mortie:
Most people don't manually deal with raw data, but a lot of people use tools like i3lock-fancy or their own imagemagick scripts to generate a lock screen image. It's not a huge change for users to modify their imagemagick-based scripts to save the result as .rgb instead of .png and provide a `--raw` option.

Now, having looked into it a bit, it seems like imagemagick dosen't support generating an ARGB image like cairo natively expects; it just supports creating an RGB or RGBA image. If it does make sense for i3lock to support raw formats, it should be able to at least convert from what imagemagick can create.

I also noticed that my error handling code has a couple of issues, like first setting `img = NULL` and then calling `cairo_surface_destroy(img)`; I just pushed a fix to my branch. I'll work on pixel format conversion and push a commit for that later too.

mortie:
I pushed a commit which adds support for 24 bits per pixel RGB.

I created a small lock script which just takes a screenshot with scrot, pixelates it with imagemagick, uses it as a lock screen image, both with and without my patch:

``` shell
#!/bin/sh
scrot /tmp/screenshot.bmp

fast() {
	dims="$(file /tmp/screenshot.bmp | awk '{print $7 "x" $9}')"
	convert /tmp/screenshot.bmp -scale 5% -scale 2000% RGB:- \
		| i3lock --raw "$dims:rgb" --image /dev/stdin
	rm -f /tmp/screenshot.bmp
}

slow() {
	convert /tmp/screenshot.bmp -scale 5% -scale 2000% /tmp/screenshot.png
	i3lock --image /tmp/screenshot.png
	rm -f /tmp/screenshot.bmp /tmp/screenshot.bmp
}

if [ "$1" = fast ]; then
	fast
else
	slow
fi
```

For me, the "slow" (i.e without this patch) option takes 0.85-0.9 seconds according to `time`, while the fast option takes 0.45-0.5 seconds. That's already a pretty noticeable difference, and will be even more noticeable with dedicated tools which can screenshot and manipulate the image without ever serializing it to the filesystem such as that i3lock-fancy-rapid fork of mine.

stapelberg:
> I somewhat agree that compared to other proposals of supporting JPEG, BMP, … this is at least a fairly basic/generic approach. At the same time I'm not sure how much this would be used as most people don't deal with raw data. @stapelberg What do you think?

I agree that supporting raw pixels is preferable over supporting additional image formats, provided we can implement this in a way where we’re reasonably certain that the input is sanitized/safe.

In this case, that would probably mean ensuring that the correct number of bytes is read (preventing out of bounds).

I briefly looked at the documentation for `cairo_image_surface_get_data` and it seems like this use-case is supported. If there are other pieces of software which do the same thing, that would be an additional boost in confidence.

In general, I’m sympathetic to the performance argument and think that providing raw image support as a building block for third-party wrappers makes sense.

mortie:
I went digging a bit for places where people are directly writing bytes to the buffer returned by `cairo_image_surface_get_data`, and found these:

* GTK modifies it when blurring: https://github.com/GNOME/gtk/blob/3f61360fa11fbc9e61585f40e197b4738f963829/gsk/gskcairoblur.c#L250
* Gecko seems to use a cairo buffer as a buffer to draw arbitrary to: https://github.com/mozilla/gecko-dev/blob/a759d6f62bdac9defd8ec7bbbef74e50474f3050/widget/gtk/nsDragService.cpp#L410

Something I should have done though, is to respect cairo's stride, not just assume that the number of bytes per line is the image width times the number of bytes per pixel. It also looks like I should probably have called `cairo_surface_mark_dirty` after drawing to it, and maybe `cairo_surface_flush` before drawing to it. I pushed changes to fix that.

---

Unrelated, but sorry for the large number of commits just to make CI happy. I should start compiling with `CFLAGS="-Wformat -Wformat-security -Wextra -Wno-unused-parameter -Werror"` like CI does. Maybe it would've been a good idea to make those warnings enabled by default in all builds?

mortie:
What's the status on this currently? I don't mean to nag at all, and understand that things take time, but I'm excited about this, and have some i3lock-using friends I'd love to geek out over locking scripts with once this is merged.

Is there anything I can do to make the process go smoother?

stapelberg:
Sorry, I was traveling recently and this fell through the cracks. I’ll try to take a look tonight.

i3lock.c
```diff
@@ -645,6 +645,113 @@ void handle_screen_resize(void) {
     redraw_screen();
 }
 
+static int read_raw_image_native(uint32_t *dest, FILE *src, int width, int height, int stride) {
+    int count = 0;
+    for (int y = 0; y < height; ++y) {
```
stapelberg: Here and below, use y++ for consistency with the code base.

i3lock.c
```diff
@@ -645,6 +645,113 @@ void handle_screen_resize(void) {
...
+    if (cairo_surface_status(img) != CAIRO_STATUS_SUCCESS) {
+        fprintf(stderr, "Could not create surface: %s\n",
+                cairo_status_to_string(cairo_surface_status(img)));
+        return NULL;
+    }
+    cairo_surface_flush(img);
+
+    /* Use uint32_t* because cairo uses native endianness */
+    uint32_t *data = (uint32_t *)cairo_image_surface_get_data(img);
+    int stride = cairo_image_surface_get_stride(img) / 4;
```
stapelberg: nit: mark this `const int`

i3lock.c
```diff
@@ -645,6 +645,113 @@ void handle_screen_resize(void) {
...
+
+static cairo_surface_t *read_raw_image(const char *image_path, const char *image_raw_format) {
+    cairo_surface_t *img;
+
+    /* Parse format as <width>x<height>[:<format>] */
+    int w = -1, h = -1;
+    char format[10];
+    strcpy(format, "native");
+    if (sscanf(image_raw_format, "%ix%i:%9s", &w, &h, format) == EOF || w == -1 || h == -1) {
+        fprintf(stderr, "Invalid image format \"%s\"\n", image_raw_format);
```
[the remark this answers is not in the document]
mortie: No, nothing in that expression sets errno; the scanf family will return EOF if there's a matching failure, if it encounters end of input before the first conversion, or if there's a read error. Errno is only set in the last case, and that doesn't apply to sscanf.

i3lock.c
```diff
@@ -645,6 +645,113 @@ void handle_screen_resize(void) {
...
+    free(buf);
+    return count;
+}
+
+static cairo_surface_t *read_raw_image(const char *image_path, const char *image_raw_format) {
+    cairo_surface_t *img;
+
+    /* Parse format as <width>x<height>[:<format>] */
+    int w = -1, h = -1;
+    char format[10];
```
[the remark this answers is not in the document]
mortie: It's 10 because the format option has to be hard coded as part of the format string for scanf, as it doesn't support `%.*s` like printf. We could do `char format[strlen("native") + 1]`, but scanf would then have to be `"%ix%i:%6s"`. I don't know that there's a nice way to fix this.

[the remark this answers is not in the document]
stapelberg: What’s wrong with the sscanf format "%ix%i:%6s"? Seems fine to me.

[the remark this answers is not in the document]
mortie: It's just that when using `char format[strlen("native")+1]`, it sort of looks like we can change the max length of a format by changing "native" to some other word, and not notice the significance of the hard-coded `%6s` on the next line. When the number is hard-coded, I think one would read the surrounding code more carefully before changing it.

In any case, I'm not extremely fond of either solution, so if you think changing it to `char format[strlen("native") + 1]` (or maybe rather `char format[sizeof("native")]` to make it a statically sized array) is better, I'm fine with that.

i3lock.1
```diff
@@ -74,6 +76,11 @@ verified or whether it is wrong).
 .BI \-i\  path \fR,\ \fB\-\-image= path
 Display the given PNG image instead of a blank screen.
 
+.TP
+.BI \-r\  format \fR,\ \fB\-\-raw= format
+Read the image given by \-\-image as a raw image instead of PNG. The argument is the image's format
+as <width>x<height>:<pixfmt>. The supported formats are 'native' and 'rgb'.
```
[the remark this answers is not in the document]
mortie: I added an example and an explanation of the pixel formats. Does that look okay?

i3lock.c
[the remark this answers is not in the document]
stapelberg: Ah, now I understand your concern. How about this, then?
```c
  char format[] = "native";
  char *fmtstring = NULL;
  if (asprintf(&fmtstring, "%%%lds", sizeof(format)-1) == -1) {
    errx(…)
  }
  sscanf(…, fmtstring, …);
  free(fmtstring);  
```

i3lock.1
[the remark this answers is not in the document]
stapelberg: Yep, this looks good. Thanks

i3lock.c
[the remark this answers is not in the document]
mortie: I pushed a commit which implements that now. However, note that asprintf is a GNU thing, not standard C.

I would personally have just gone with using a bigger than necessary fixed sized array (maybe a nice round size like 16 rather than my arbitrary 10) instead of relying on non-standard APIs which allocate memory for such a small thing, but I also completely understand the desire to not have sizes hard-coded like that.

i3lock.c
```diff
@@ -645,6 +645,119 @@ void handle_screen_resize(void) {
...
+    int w = -1, h = -1;
+    char format[] = "native";
+    char *fmtstring = NULL;
+    if (asprintf(&fmtstring, "%%ix%%i:%%%lds", sizeof(format) - 1) == -1) {
+        errx(EXIT_FAILURE, "asprintf() failed.");
+    }
+
+    if (sscanf(image_raw_format, fmtstring, &w, &h, format) == EOF || w == -1 || h == -1) {
+        fprintf(stderr, "Invalid image format \"%s\"\n", image_raw_format);
+        return NULL;
```
stapelberg: Add a `free(fmtstring);` here to prevent a memory leak.

stapelberg:
Also, can you squash your commits please? We’d like this to be one logical git commit in the history after merging.

stapelberg:
@psychon Would you have a second to take a look at this PR please? In particular, I’m hoping you might be familiar with how cairo images are laid out in memory, and whether this code gets it right regarding striping etc. — I trust the author has verified functionality, so things seem correct, but perhaps there’s some nuance to it. Thanks in advance,

mortie:
Commits squashed.

> I still don’t understand why you would use a bigger-than-necessary array? Why not size it exactly as large as it needs to be?

I don't know, mostly habit. I think of it like a buffer which has to be at least a certain size; making it a bit bigger than that doesn't hurt, decreases the chances of accidentally making the buffer slightly too small (either immediately or through changes in the future), and [compilers tend to do a bunch of alignment anyways](https://godbolt.org/z/xqOhrk).

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
...
+        count += n;
+        if (n < (size_t)(width * 3))
+            break;
+
+        for (int x = 0; x < width; ++x) {
+            int idx = x * 3;
+            dest[y * stride + x] = 0 |
+                                   (buf[idx + 0] << 16) |
+                                   (buf[idx + 1] << 8) |
+                                   (buf[idx + 2]);
```
psychon: Why are "native" images in native endian, but "rgb" images are always in little endian (well, or rather "byte by byte" since 24 bits is a weird word size)? That is a bit surprising to me (but I did not read the man page diff carefully and it is documented as such).

I am not really sure what the right course of action is here, but I bet that someone will get the endianess with "native" wrong. How about making the two "channel orders" the same on big endian instead of little endian, so that people on little endian notice that the order is off and *hopefully* do the right thing? Or... how about basing both formats on bytes instead of the machine word like cairo does?

i3lock.1
```diff
@@ -74,6 +76,15 @@ verified or whether it is wrong).
 .BI \-i\  path \fR,\ \fB\-\-image= path
 Display the given PNG image instead of a blank screen.
 
+.TP
+.BI \-r\  format \fR,\ \fB\-\-raw= format
+Read the image given by \-\-image as a raw image instead of PNG. The argument is the image's format
+as <width>x<height>:<pixfmt>. The supported pixel formats are 'native' and 'rgb'.
+The "rgb" pixel format expects a pixel to be three bytes; red, green, and blue.
+The "native" pixel format expects a pixel to be ARGB, as a 32-bit (4-byte) integer in
+the machine's native endianness. The alpha is ignored.
```
psychon: Why does it need to be called "alpha" if it is ignored?

I am asking, because cairo uses pre-multiplied alpha: Red with 50% transparency has value 0x7f in the red component, not 0xff. If you just do not mention it as alpha, this problem "goes away".

Here is the doc for `CAIRO_FORMAT_RGB24`, for "wording inspiration":

> each pixel is a 32-bit quantity, with the upper 8 bits unused. Red, Green, and Blue are stored in the remaining 24 bits in that order.

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
     redraw_screen();
 }
 
+static int read_raw_image_native(uint32_t *dest, FILE *src, int width, int height, int stride) {
+    int count = 0;
+    for (int y = 0; y < height; y++) {
+        size_t n = fread(&dest[y * stride], 1, width * 4, src);
+        count += n;
+        if (n < (size_t)(width * 4))
```
psychon: Does this code have to worry about integer overflows? (I'll only leave this comment here once, but want to point out that `y * stride` could also overflow and could instead be done via `dest += stride / 4`, which would however "look funny")

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
...
+}
+
+static cairo_surface_t *read_raw_image(const char *image_path, const char *image_raw_format) {
+    cairo_surface_t *img;
+
+    /* Parse format as <width>x<height>[:<format>] */
+    int w = -1, h = -1;
+    char format[] = "native";
+    char *fmtstring = NULL;
+    if (asprintf(&fmtstring, "%%ix%%i:%%%lds", sizeof(format) - 1) == -1) {
```
psychon: I noticed that there was some discussion regarding this, but... really?

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
...
+        errx(EXIT_FAILURE, "asprintf() failed.");
+    }
+
+    if (sscanf(image_raw_format, fmtstring, &w, &h, format) == EOF || w == -1 || h == -1) {
```
psychon: What is so special about -1 that I cannot claim my image has width/height -1, but -2 is fine? :-P
(Yes, I know that this is meant to check for "did `sscanf` modify the argument at all?")

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
...
+        fprintf(stderr, "Invalid image format \"%s\"\n", image_raw_format);
```
psychon: Hm... I cannot add two separate comments to the same line? Anyway, the call to `sscanf` is still wrong:

```
RETURN VALUE
       On  success,  these functions return the number of input items successfully matched and assigned;
       this can be fewer than provided for, or even zero, in the event of an early matching failure.
```
Or is this why you check against `-1` instead of checking for `sscanf` to return 3?

Also... doesn't gcc warn against non-static format strings?

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
...
+    if (cairo_surface_status(img) != CAIRO_STATUS_SUCCESS) {
+        fprintf(stderr, "Could not create surface: %s\n",
+                cairo_status_to_string(cairo_surface_status(img)));
+        return NULL;
+    }
+    cairo_surface_flush(img);
+
+    /* Use uint32_t* because cairo uses native endianness */
+    uint32_t *data = (uint32_t *)cairo_image_surface_get_data(img);
+    const int stride = cairo_image_surface_get_stride(img) / 4;
```
psychon: Uhm. It surprises me that you and cairo use different definitions of "stride". I know that this has to be divided by four due to C, but... still, this surprises me here.

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
...
+    }
+
+    /* Read the image, respecting cairo's stride, according to the format */
+    int size, count;
+    if (strcmp(format, "native") == 0) {
+        /* If the format is 'native', just read each line directly into the buffer */
+        size = w * h * 4;
+        count = read_raw_image_native(data, f, w, h, stride);
+    } else if (strcmp(format, "rgb") == 0) {
+        /* If the format is 'rgb', we have to convert it to argb */
```
psychon: I wouldn't really call this ARGB. You set the "alpha" byte to zero, not 0xff.

i3lock.c
```diff
@@ -645,6 +645,120 @@ void handle_screen_resize(void) {
...
+            return NULL;
+        } else {
+            /* Print a warning if the file contains less data than expected,
+             * but don't abort. It's useful to see how the image looks even if it's wrong. */
+            fprintf(stderr, "Warning: expected to read %i bytes from \"%s\", read %i\n",
+                    size, image_path, count);
+        }
+    }
+
+    fclose(f);
```
psychon: I didn't check if any of the error paths leak something. This is complicated enough that I would want to double check, but I am tired enough that I do not want to do this. I'll just assume someone double checked this.

i3lock.c
mortie: The idea is that "native" is as close to what cairo expects as possible, to achieve the lowest overhead possible (I would probably have just used one big fread() and made the stride part of the public interface, if how cairo calculates the stride is part of its public interface, but cairo doesn't document how it calculates strides afaik, and I didn't want to rely on an implementation detail), and is meant to be used by tools; "rgb" on the other hand is there just as a convenience because RGB, with the red byte first, green byte second, and blue byte last, is the most common user-facing raw format in my experience. For example, imagemagick can output the kind of RGB this patch implements, but you can't exactly ask it for 32-bit pixels in native endianness where the highest 8 bits are ignored.

Any change would, in my opinion, have to preserve both the "native" format (for dedicated tools) and the "rgb" format (for user-created scripts), so in my opinion, the options are leaving it as it is, or adding a native-endian rgb24 version (and possibly a big-endian rgb32 version?).

i3lock.c
mortie: If we were using size_t I think we should be fine; according to my math, given big enough screens (say, a 64k screen) we would come a bit close to overflowing a 32-bit size_t, but we'd also be coming a bit close to exhausting all of the 32-bit address space given that everything is read into RAM. On a 64-bit system it should be fine.

However, since both y and stride are ints, it can be a 32-bit multiplication on even 64-bit systems. We should probably make sure all the relevant operations are with size_t operands.

i3lock.c
mortie: We check w and h separately because those are the only required options, the format is optional (and defaults to native). It would be possible to check that sscanf returns at least 2, but I don't think it's a big deal. It would probably make sense though to check if w or h are less than 0, rather than checking if they're exactly -1, as you pointed out in the other comment.

i3lock.c
stapelberg: We couldn’t find a way to express `strlen("native")` at compile time within a format string. Do you know of one?

i3lock.c
stapelberg: I think the >= 2 check actually sounds better than checking an individual error condition.

i3lock.c
stapelberg: All returns are preceded by an fclose(f), so that looks good to me.

i3lock.c
mortie: I renamed it to pixstride to make it less confusing. That seems to be what cairo internally uses when talking about a stride in terms of pixels rather than bytes.

mortie:
Currently, the default pixel format is "native". I'm not sure if that's correct, or if it would make more sense for "rgb" to be the default. What do you people think?

"native" sort of makes sense as a default, because it's the format which doesn't loop through each pixel in order to convert it. However, at the same time, people writing their own commands are more likely to use "rgb" from imagemagick or some other tool, while the people using "native" will mostly be people who write wrappers around i3lock, where the one who's writing the wrapper only has to write the i3lock command once, and can be expected to know more about how pixel formats work; if rgb was the default, only the people writing tools would have to specify anything.

Maybe the best way is to just not default to anything, to always require either rgb or native?

owenthewizard:
IMO it's more realistic to expect people writing wrappers to worry about specifying `native` and default to `rgb` for the end user's sake. Not having a default sounds good too.

stapelberg:
+1 for being explicit (no default). This is an advanced option, and spelling out all the details results in more clarity.

i3lock.c
psychon: > I would probably have just used one big fread() and made the stride part of the public interface, if how cairo calculates the stride is part of its public interface, but cairo doesn't document how it calculates strides afaik, and I didn't want to rely on an implementation detail)

I would expect lots of breakage if cairo changes it requirements here, so personally, I would be fine with assuming the current stride logic to stay the way it is. For RGB24 and ARGB32, the stride logic is: There is no padding. That's especially nice to implement. ;-)

(The technical definition is: Let `bpp` be the bits per pixel of the selected format. The stride is `ceil(bpp*width/8)` rounded up to the next multiple of four. For RGB24 and ARGB32, `bpp=32` and the formula simplifies to `width*4`, i.e. just four bytes per pixel. This aligned-to-multiple-of-four is required by pixman to be able to use e.g. SSE.)

i3lock.c
psychon: Not really. I guess I would do it the other way around. I feel like that way might be easier to understand. (50 is just some number larger than `strlen("native")`)
```C
#define MY_LENGTH 50
#define STRINGIFY1(x) #x
#define STRINGIFY(x) STRINGIFY1(x)
	char buffer[MY_LENGTH+1];
	sscanf(string, "%" STRINGIFY(MY_LENGTH) "s", buffer);
```

mortie: I believe this is essentially the last remaining point before I squash commits again and things are ready to merge (unless there are any issues I haven't noticed with my new changes of course). I don't particularly care if we leave it as it is, revert to how I originally had it, or do something like what psychon suggests (which is fairly similar to how it was originally, except that the length is set by a macro).

stapelberg: Let’s go with @psychon’s suggestion, but using a buffer size that is only as large as we need.

mortie: Alright, I changed it: https://github.com/i3/i3lock/pull/226/commits/80e6215e6f45e4615f11558bcc13e2a04f4f880a

i3lock.c
mortie: I still think it's okay to just deal with stride as the code currently does; it doesn't hurt to be on the safe side in case cairo does actually change their stride calculations, and it shouldn't really have an effect on performance.

Can this conversation be resolved, or do you disagree?

stapelberg:
This seems good enough for now; we can always fix things later if required.

Thanks for your work on this!

mortie:
Hmm, I noticed earlier today that i3lock-color, the seemingly fairly popular fork of i3lock, already uses the -r short opt. Since this feature also uses -r, that project would need to either break backwards compat, or change this feature to use a different option instead (such as -R). Does that affect amything? To what extent should i3lock try to not add options which conflict with popular forks?

It would be a quick fix to change the short opt from -r to something else if you think that's warranted.

stapelberg:
Given that it’s a quick fix, let’s change it.
We could also just not give a short option for this at all, and just use `--raw`, which is more descriptive anyway (and you’re not typing i3lock commands by hand anyways).

owenthewizard:
Could an `rgba` format be added, identical to `rgb` but with an extra byte (that is ignored)? This would allow for easier compatibility with XCb screenshots.

stapelberg:
Isn’t that what the native format does?

mortie:
No, native is ARGB (or, well, XRGB, since the "alpha" byte is ignored), and in the machine's endianness (so since most machines are little endian, that usually means the actual bytes are stored as BGRX).

I took some time to add support for the XRGB, RGBX, BGR, XBGR, and BGRX: https://github.com/mortie/i3lock/commit/9532412db61abac45be31bd328849bac320f4561. What do you think @stapelberg, shall I submit a pull request?

stapelberg:
Do we really need all the variants? Are they really all common? But yeah, submit a PR please.

mortie:
The advantage to just supporting a lot of different formats is that it makes wasteful pixel formats conversions less likely. Maybe someone, say, uses a library which returns a buffer of native-endian RGBA data; they would want to use the XBGR format on little-endian systems and RGBX on big-endian systems. I have worked a lot with raw images, and can say that I have been extremely happy more than once for libraries providing support for weird formats (such as libYUV's NV12ToABGR).

 It isn't really any more difficult to support all those formats than to just support a few of them; each format other than native is just a different configuration of how many bytes per pixel there are and what order they are in.

Another completely valid stance would be to say that i3lock just supports the "native" format for anyone who wants maximum performance and the "rgb" format for anyone who wants to use imagemagick or other command-line tools which produce raw images, and say that cases such as xcb screenshots should be solved by just converting from RGBA to "native" as the data is written to a file (or i3lock's stdin). That wouldn't really be any more or less efficient, it would just make i3lock sightly smaller and make it a bit harder to make a tool which writes raw image data to i3lock.

In any case, pull request submitted.

PR #230, opened 2019-04-25
Added support for more image types and automatic resizing

diogo464:
Added support to read more images using stb_image header. Images are automatically resized to fit the screen but can also be used with original size.


Airblader:
Sorry, we're not looking to support more image formats. Our current idea is to go with #226.

PR #245, opened 2019-08-09
WIP: add crude config file support

granquet:
Signed-off-by: Guillaume Ranquet <ranquet.guillaume@gmail.com>

This is a very crude implementation of a config file support...
Pull request mainly for discussion.
if upstream is interested in the feature, I'll rework the patch to have something a bit more robust.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
     }
 }
 
+struct option longopts[] = {
+    {"version", no_argument, NULL, 'v'},
+    {"nofork", no_argument, NULL, 'n'},
+    {"beep", no_argument, NULL, 'b'},
+    {"dpms", no_argument, NULL, 'd'},
+    {"color", required_argument, NULL, 'c'},
+    {"pointer", required_argument, NULL, 'p'},
+    {"debug", no_argument, NULL, 0},
+    {"help", no_argument, NULL, 'h'},
+    {"no-unlock-indicator", no_argument, NULL, 'u'},
+    {"image", required_argument, NULL, 'i'},
+    {"raw", required_argument, NULL, 0},
+    {"tiling", no_argument, NULL, 't'},
+    {"ignore-empty-password", no_argument, NULL, 'e'},
+    {"inactivity-timeout", required_argument, NULL, 'I'},
+    {"show-failed-attempts", no_argument, NULL, 'f'},
+    {NULL, no_argument, NULL, 0}};
+
+
+char *image_path = NULL;
+char *image_raw_format = NULL;
+int curs_choice = CURS_NONE;
+
+int set_opt(char o, char *optarg, int longoptind) {
+    switch (o) {
+        case 'v':
+            errx(EXIT_SUCCESS, "version " I3LOCK_VERSION " © 2010 Michael Stapelberg");
+        case 'n':
+            dont_fork = true;
+            break;
+        case 'b':
+            beep = true;
+            break;
+        case 'd':
+            fprintf(stderr, "DPMS support has been removed from i3lock. Please see the manpage i3lock(1).\n");
+            break;
+        case 'I': {
+                      fprintf(stderr, "Inactivity timeout only makes sense with DPMS, which was removed. Please see the manpage i3lock(1).\n");
+                      break;
+                  }
+        case 'c': {
+                      char *arg = optarg;
+
+                      /* Skip # if present */
+                      if (arg[0] == '#')
+                          arg++;
+
+                      if (strlen(arg) != 6 || sscanf(arg, "%06[0-9a-fA-F]", color) != 1)
+                          errx(EXIT_FAILURE, "color is invalid, it must be given in 3-byte hexadecimal format: rrggbb");
+
+                      break;
+                  }
+        case 'u':
+                  unlock_indicator = false;
+                  break;
+        case 'i':
+                  image_path = strdup(optarg);
+                  break;
+        case 't':
+                  tile = true;
+                  break;
+        case 'p':
+                  if (!strcmp(optarg, "win")) {
+                      curs_choice = CURS_WIN;
+                  } else if (!strcmp(optarg, "default")) {
+                      curs_choice = CURS_DEFAULT;
+                  } else {
+                      errx(EXIT_FAILURE, "i3lock: Invalid pointer type given. Expected one of \"win\" or \"default\".");
+                  }
+                  break;
+        case 'e':
+                  ignore_empty_password = true;
+                  break;
+        case 0:
+                  if (strcmp(longopts[longoptind].name, "debug") == 0)
+                      debug_mode = true;
+                  else if (strcmp(longopts[longoptind].name, "raw") == 0)
+                      image_raw_format = strdup(optarg);
+                  break;
+        case 'f':
+                  show_failed_attempts = true;
+                  break;
+        default:
+                  errx(EXIT_FAILURE, "Syntax: i3lock [-v] [-n] [-b] [-d] [-c color] [-u] [-p win|default]"
+                          " [-i image.png] [-t] [-e] [-I timeout] [-f]");
+    }
+
+    return 0;
+}
+
+int conf_file_rc() {
+
+    char buff[128] = {0};
+    char path[1024] = {0};
+    char *p = getenv("HOME");
+    int i = 0;
+
+    if(p == NULL)
+        return -ENOENT;
+
+    strcat(path, p);
+    strcat(path, "/.i3lockrc");
+    FILE *f = fopen(path, "r");
+
+    if(f == NULL)
+        return -ENOENT;
+
+    while(fgets(buff, sizeof buff, f) != NULL) {
+
+        for(i=0; i<(sizeof longopts / sizeof *longopts) - 1; i++) {
+            if(strncmp(longopts[i].name, buff, strlen(longopts[i].name)) == 0) {
+                if((p = strchr(buff, '\n')) != NULL)
+                        *p = '\0';
+                set_opt(longopts[i].val, buff + strlen(longopts[i].name)+1, i);
+                break;
+            }
+        }
+    }
+
+    return 0;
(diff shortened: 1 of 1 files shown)


Airblader:
> Pull request mainly for discussion.

Fair enough, but before we can start that I would ask you to explain a bit as to why you would like to see config files for i3lock as that's the main question to answer, right? :-)

One reason I personally could see is that it would allow shipping things like systemd unit files while leaving configuration to the user, although in this case we'd probably want to have behavior such as command line arguments always overriding the config file.

granquet:
Oh... mainly because I'm lazy :)

I'm running dmenu and call i3lock from dmenu.
and as you probably know, dmenu doesn't handle aliases... which means I'm always locking my screen with that white default screen...

I could run i3lock from a shell prompt, but most of the times I'm ssh'ing somewhere or running inside a tmux session with DISPLAY unset (because I started the session from another machine using ssh usually).

I could also create a script and move it in my $PATH...

I like the config file better, it's easier to bring with me when I'm getting a new account on another machine.

this is already coded so that the config file is always overriden by cmdline arguments as the config_file is parsed before the call to getopt_long().

ivanhercaz:
A configuration file to save time configuring another computer or make easier to have a global/general configuration would be very useful. @Airblader, would it be possible to apply this to i3lock?

I don't know very much about C, but if I am understanding fine the check done by Travis, the problem is in the format of the new code, specifically with the indentation. Am I right? I am going to open a review to point out these mistakes.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+            break;
+        case 'b':
+            beep = true;
+            break;
+        case 'd':
+            fprintf(stderr, "DPMS support has been removed from i3lock. Please see the manpage i3lock(1).\n");
+            break;
+        case 'I': {
+                      fprintf(stderr, "Inactivity timeout only makes sense with DPMS, which was removed. Please see the manpage i3lock(1).\n");
+                      break;
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                  }
+        case 'c': {
+                      char *arg = optarg;
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                      if (arg[0] == '#')
+                          arg++;
+
+                      if (strlen(arg) != 6 || sscanf(arg, "%06[0-9a-fA-F]", color) != 1)
+                          errx(EXIT_FAILURE, "color is invalid, it must be given in 3-byte hexadecimal format: rrggbb");
+
+                      break;
+                  }
+        case 'u':
+                  unlock_indicator = false;
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                  break;
+        case 'i':
+                  image_path = strdup(optarg);
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                  break;
+        case 't':
+                  tile = true;
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                  break;
+        case 'p':
+                  if (!strcmp(optarg, "win")) {
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                      curs_choice = CURS_WIN;
+                  } else if (!strcmp(optarg, "default")) {
+                      curs_choice = CURS_DEFAULT;
+                  } else {
+                      errx(EXIT_FAILURE, "i3lock: Invalid pointer type given. Expected one of \"win\" or \"default\".");
+                  }
+                  break;
+        case 'e':
+                  ignore_empty_password = true;
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                  break;
+        case 0:
+                  if (strcmp(longopts[longoptind].name, "debug") == 0)
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                      debug_mode = true;
+                  else if (strcmp(longopts[longoptind].name, "raw") == 0)
+                      image_raw_format = strdup(optarg);
+                  break;
+        case 'f':
+                  show_failed_attempts = true;
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                  break;
+        default:
+                  errx(EXIT_FAILURE, "Syntax: i3lock [-v] [-n] [-b] [-d] [-c color] [-u] [-p win|default]"
```
ivanhercaz: Indentation error, should be like in `case 'd':`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+}
+
+int conf_file_rc() {
+
+    char buff[128] = {0};
+    char path[1024] = {0};
+    char *p = getenv("HOME");
+    int i = 0;
+
+    if(p == NULL)
```
ivanhercaz: `if (p == NULL)`

Needs a space.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+        return -ENOENT;
+
+    strcat(path, p);
+    strcat(path, "/.i3lockrc");
+    FILE *f = fopen(path, "r");
+
+    if(f == NULL)
```
ivanhercaz: `if (f == NULL)`

Needs a space.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+        return -ENOENT;
+
+    while(fgets(buff, sizeof buff, f) != NULL) {
```
ivanhercaz: `while (fgets(buff, sizeof buff, f) != NULL)`

The `while` needs a space before the condition.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+
```
ivanhercaz: This newline should be removed.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+        for(i=0; i<(sizeof longopts / sizeof *longopts) - 1; i++) {
```
ivanhercaz: The `for` loop needs a space before `(i=0; i<(sizeof longopts / sizeof *longopts) - 1; i++)` and this line needs the next changes:
  - `i=0` » `i = 0`
  - `i<(sizeof longopts / sizeof *longopts)` » `i < (sizeof longopts / sizeof *longopts)`

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+            if(strncmp(longopts[i].name, buff, strlen(longopts[i].name)) == 0) {
```
ivanhercaz: The `if` needs a space before `(strncmp(...`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                if((p = strchr(buff, '\n')) != NULL)
```
ivanhercaz: The `if` needs a space before `((p = strchr(...`.

i3lock.c
```diff
@@ -1004,105 +1004,148 @@ static void raise_loop(xcb_window_t window) {
...
+                        *p = '\0';
+                set_opt(longopts[i].val, buff + strlen(longopts[i].name)+1, i);
```
ivanhercaz: `strlen(longopts[i].name)+1` should be `strlen(longopts[i].name) + 1`.

granquet:
@ivanhercaz : I'll update / cleanify the code and make it ready for the review / merge process as soon as @Airblader agrees the feature is indeed worth working on.

besides indentation, the code might be missing some checks here and there to make it robust and fail graciously whenever there's an issue.

stapelberg:
We definitely shouldn’t add an ad-hoc config file parser; if we added anything, it would need to be using libconfuse just like i3status (rather than the custom parser i3 itself uses).

That said, I don’t think it’s worthwhile to have config file support for i3lock. One typically sets very few options, if any, and specifying them via flags works fine.

The cited use-case of interactively launching i3lock is better handled by starting a wrapper script or binding a key in i3 instead of using a program launcher.

granquet:
> The cited use-case of interactively launching i3lock is better handled by starting a wrapper script or binding a key in i3 instead of using a program launcher.

let's agree to disagree on that one :)
I don't want another key binding to remember and I don't want to fiddle with my $PATH for one wrapper script since dmenu doesn't recognize aliases.

that's nothing I cannot live with... I just run i3lock with its defaults and it works, but that's not what I *really* want.

I'm still willing to work on the feature if there's any interest. feel free to re-open the merge request if you change your mind.

PR #247, opened 2019-09-20
Add a note about the image processing issues

ivanhercaz:
As it is indicated in #37, #62, #66 and #85, i3lock doesn't planned to add this feature. So I think it is a good idea to write a note about it in the second point of the introduction, dedicated to the background settings.

@Airblader, do you think I should add links to the issues/PRs closed due to this? If they aren't necessary and the team is agree with this change, this PR is fine and complete.

Regards,
Iván

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ Many little improvements have been made to i3lock over time:
    after waking up your computer from suspend to RAM)
 
 - You can specify either a background color or a PNG image which will be
-  displayed while your screen is locked.
+  displayed while your screen is locked (**note**: i3lock don't make any
+  image manipulation or processing, nor it is planned to add this feature).
 
 - You can specify whether i3lock should bell upon a wrong password.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,8 @@ Many little improvements have been made to i3lock over time:
    after waking up your computer from suspend to RAM)
 
 - You can specify either a background color or a PNG image which will be
-  displayed while your screen is locked.
+  displayed while your screen is locked (**note**: i3lock don't make any
```
Airblader: Can we rephrase this a little bit?

> […] while your screen is locked. Note that i3lock is not an image manipulation software. If you need to resize the image to fill the screen or similar, use existing tooling to do this before passing it to i3lock.

ivanhercaz: Of course! I will perform the suggested change.

Airblader:
Thanks!

PR #250, opened 2019-10-08
Single keypress grace period

kosmisk-dk:
A short configurable period in seconds (default: disabled/0). Where a single keypress is enough to unlock. For all practical purposes postpone the locking of the screen, and only blanking it.


Airblader:
Thanks for the PR. This is a duplicate of #158, so I will refer you there for a reasoning of why we do not want to add this to i3lock. Thanks for understanding!

PR #266, opened 2020-03-04
Add a no background option

loki7990:
I.e., Use nothing if the user doesn't use a color/image/blurred background. I'm requesting this because I currently use a script to make cmatrix go fullscreen just before locking the screen. Would be nice to be able to see it while the screen is locked.


Airblader:
Sorry, but transparent lockscreens is a footgun we'd like to stay clear from.

PR #267, opened 2020-04-18
restore DPMS ability (turn off screen when locked)

wheelcomplex:
This is a PoC just for reference, most code ware copyed from https://github.com/i3/i3lock/commit/929e541fb78aa167fc52eb684d4fdff8b305b246 

Tested on Ubuntu 20.04 only.


orestisfl:
Since there are no plans to restore this AFAIK, I am closing this PR.

PR #268, opened 2020-04-19
"SortIncludes:  false " for clang-format-10

wheelcomplex:
set "SortIncludes:    false " explicitly for clang-format-10

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -8,3 +8,4 @@ IndentWidth: 4
 PointerBindsToType: false
 ColumnLimit: 0
 SpaceBeforeParens: ControlStatements
+SortIncludes:    false


stapelberg:
Thanks for your PR.

What’s the motivation for making this change? You’re only describing what you’re doing, not why.

wheelcomplex:
> Thanks for your PR.
> 
> What’s the motivation for making this change? You’re only describing what you’re doing, not why.

Oops, because clang-format-10 (default in Ubuntu 20.04, and clang-format-3.5 is not available) will re-sort includes (and break CI) without this.

Unfortunately SortIncludes is not supported by clang-format-3.5.

<pre>
$ clang-format-3.5 -i *.[ch] && git diff --exit-code || (echo 'Code was not formatted using clang-format!'; false)
YAML:11:18: error: unknown key 'SortIncludes'
SortIncludes:    false
                 ^~~~~
</pre>

stapelberg:
We should probably just upgrade to a newer version, as we have already done for the i3 repository itself (clang-format-6.0).

PR #271, opened 2020-05-20
Add option to remove text indicators

desmondcheongzx:
Add an additional option -r/--remove-text-indicator to disable
the text indicator when locking or unlocking.
i.e. when it says "Verifying..." or "Wrong!".

This is an option for a minor aesthetic change, but I've made
multiple tweaks such as this in order to customize my desktop
experience and I wonder if others might find this useful too.


stapelberg:
Thanks for your PR, but I’ll have to decline it. We want to keep i3lock as small as possible, and not necessarily cater to everyone’s subjective preferences.

Issue #115, opened 2017-02-13
i3lock fails within a second if mouse cursor is grabbed/occupied


GSI:
Since tools like **unclutter** grab the mouse cursor, **i3lock** will fail. The bigger problem is that it doesn't fail directly - it shows the lock screen for a second and then, when you already left the workplace, your session is unlocked.

I work around the issue using the following function:

```
lock_screen () {
        DISPLAY_GEOMETRY=$(xdotool getdisplaygeometry | sed 's# #x#g')

        # NOTE mouse is moved as the lock operation otherwise fails:
        xdotool mousemove 0 0 && xdotool mousemove --polar 0 0 &&
                i3lock -bd
}
```

**However** this doesn't liberate the cursor in case tools like **xkill** occupy it.

Is there any reliable - and thus secure - way to handle such situations?

Why does i3lock even care about a mouse cursor? Can this "dependency" be removed?

Airblader:
> The bigger problem is that it doesn't fail directly - it shows the lock screen for a second and then, when you already left the workplace, your session is unlocked

This isn't quite true anymore, see #88 and #99. That's really all we can do due to how X works. Given that this duplicates a few issues and we've addressed it now, I'm closing this. 

By the way, I've written unclutter-xfixes a little while ago. It's unclutter, but won't suffer from these problems.

GSI:
I just tried the most recent i3lock version from git. It delays even more before it's becomes obvious that locking failed.

Does X provide no manner to check if mouse is occupied/available?
At best we could find out the PID that it blocking it (and kill the according process).

Airblader:
> Does X provide no manner to check if mouse is occupied/available?

No, you just notice that it doesn't work.

> and kill the according process

We most certainly do not want to kill other processes. What if it's unsaved hours of work that you'd be destroying?

> I just tried the most recent i3lock version from git. It delays even more before it's becomes obvious that locking failed.

How do you start i3lock? You should see a message that locking is in progress (as you can see in the referenced PRs). Of course this assumes you actually display the unlock indicator.

GSI:
> We most certainly do not want to kill other processes. What if it's unsaved hours of work that you'd be destroying?

I'd certainly prefer loosing hours of work over loosing years of work.

Of course that would be an optional switch in i3lock or even just something in a wrapper script.

Like even if i3lock just gave the PID as exit code, the situation could be handled according to user preference.


> How do you start i3lock?

[...]

PR #272, opened 2020-06-02
Fix update keyboard layout indicator on change.

andremarcais:
This is a fix to issue #115.


andremarcais:
Oops. Wrong repository. This is for i3lock-color. Sorry.

PR #278, opened 2020-07-05
add extension support

M1cha:
This is just a proof of concept to start the conversation about weather you want this or not.
This is a small change to allow people to draw their own background using cairo while keeping their code out of i3lock. The code is loaded via libdl.

The extensions can even add their own callbacks(like timers) to the event loop without any additional i3lock logic because i3lock uses `EV_DEFAULT` anyway.

This would allow implementing things like #7 or #47 without touching i3locks code.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,6 +74,8 @@ AC_SEARCH_LIBS([floor], [m], , [AC_MSG_FAILURE([cannot find the required floor()
 # libev does not ship with a pkg-config file :(.
 AC_SEARCH_LIBS([ev_run], [ev], , [AC_MSG_FAILURE([cannot find the required ev_run() function despite trying to link with -lev])])
 
+AC_SEARCH_LIBS([dlopen], [dl])
+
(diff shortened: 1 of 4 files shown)


stapelberg:
Thanks for your PR.

I have to say I’m not keen on supporting dynamic loading in security sensitive software :-/

If people want their own look & feel so badly, they should maintain a (local?) fork of i3lock, I think.

M1cha:
Well as long as people don't load an extension, there's no security impact.
If they do, they're on their own - which we could make clear in several places.

stapelberg:
People don’t necessarily read the documentation. Rather than handing them a footgun with a disclaimer, I’d rather we didn’t have any footguns.

M1cha:
In that case the only way out would be putting the rendering code into a separate process.
Maybe just an imagebuffer allocated using shm? It would probably need two buffers and a lock to switch between them while drawing.

stapelberg:
No, shared buffers and synchronization doesn’t sound appealing either.
It’s extra complexity for a feature we don’t want to support.

PR #282, opened 2020-07-12
Enhance clear_password_memory() / explicit_bzero()

michaelortmann:
clear_password_memory() uses explicit_bzero() under openbsd and handcrafted code if not. This patch enhances clear_password_memory() to detect and use explicit_bzero(), which is also available under linux since glibc 2.25, explicit_memset(), memset_s() and a fallback. The new code is "stolen" from the great libsodium.


i3lock.c
```diff
@@ -169,29 +174,6 @@ static bool load_compose_table(const char *locale) {
...
-    volatile char *vpassword = password;
-    for (size_t c = 0; c < sizeof(password); c++)
-        /* We store a non-random pattern which consists of the (irrelevant)
-         * index plus (!) the value of the beep variable. This prevents the
-         * compiler from optimizing the calls away, since the value of 'beep'
-         * is not known at compile-time. */
-        vpassword[c] = c + (int)beep;
-#endif
-}
-
```
orestisfl, line 194: I might be missing something here but aren't you supposed to call `explicit_bzero()` here?

stapelberg:
Thanks for your pull request!

I appreciate the improvement of using `explicit_bzero` whereever available; we should definitely do that.

However, I think the value we get from the rest of the changes does not outweigh the inconvenience that a separate license attribution constitutes: downstream distributions now need to review licenses and possibly adjust their metadata.

Could you change your PR to only contain the `explicit_bzero` improvement please?

i3lock.c
stapelberg: The call moved into the define above: `#define clear_password_memory() explicit_bzero(password, strlen(password))`

orestisfl: Oh I see, I didn't notice we were in a function.

michaelortmann:
Sure, i removed the license text. Now you can add your own / whatever you want. The remaining ISC license in explicit_bzero.c shouldn't be a problem, for it allows anything (ISC is functionally equivalent to the simplified BSD and MIT licenses), as long as you keep that small copyright notice / license text in there. We could also remove the link to https://raw.githubusercontent.com/jedisct1/libsodium/d47ded1867af69965b2374b8fb90aee01e6ff291/src/libsodium/sodium/utils.c, but i thought its always a good idea, to document where you have some code from, so that you could later sync with its upstream, should there be any changes.

stapelberg:
I meant remove the licensed code altogether, not just remove the attribution :)

michaelortmann:
this licensed code _is the_ "explicit_bzero improvement". you want me to reimplement this? i didn't reuse that piece of code from libsodium light heartly. i didn't copy/paste it unmindfully like some javascript coders *sry* do. i did so because i don't  think we should mess with encryption/security stuff if we dont have to but use a well trusted source in this domain. also i always look at the license, i would never have chosen gpl or some "complicated" license, but that ISC is one of the least strict licenses. less strict than the bsd license i3lock is under. plus, i dont think one can easily replace the licensed code, because its quite to the point.

stapelberg:
> PR if you want. where we simply check for explicit_bzero via configure script instead of the hardcoded openssl-only usage

(nit: s/openssl/OpenBSD/)

Yes, this is what I meant. I think that’s where most value will be.

The libsodium code is nice, but e.g. its windows-specific code path, inline assembler and memset_explicit paths will neither be compiled nor used anyway, so let’s not use it.

michaelortmann:
Here you are. BTW: i3lock.c included <string.h> twice, so i removed one of those includes.

stapelberg:
Thanks!

PR #296, opened 2021-01-26
Update i3lock.c

prismspecs:
black lives matter


Airblader:
I know that @stapelberg supported BLM with donations given to the i3 project and he may override me on this, but I don't think this has a place in the source code like this.

PR #299, opened 2021-03-23
refresh image upon receiving USR1

mfontani:
Hi,

I've been using these proposed two patches on top of i3lock for quite some time now, and I'm testing the waters about whether they could be merged upstream instead of sitting in a branch, with no other user than myself.

The point of those patches is to allow refreshing the image used for locking the screen upon receiving a signal (i.e. `USR1`).

As to the use-case...

When one wants to use an amended screenshot of the current screen to lock their screen, they usually proceed as follows:

1. take screenshot
2. manipulate screenshot image
3. call i3lock with the manipulated image

The step at "2." may take quite a bit of time, and while "2." performs its tasks *the screen isn't yet locked*. If "2." moreover fails, it's possible the screen is never effectively locked.

With these patches, my process is now:

1. take screenshot
2. call i3lock with the just-screenshotted image (*quick*, and the *screen is now locked*)
3. manipulate screenshot image (may take however long, screen is still locked)
4. pkill -USR1 i3lock, which refreshes the image (screen is still locked, but now displays the new image)

In case the "3." step above takes an inordinate amount of time or fails, the screen is locked.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -97,6 +97,9 @@ bool tile = false;
 bool ignore_empty_password = false;
 bool skip_repeated_empty_password = false;
 
+/* Global, to allow it to be refreshed by signal handler */
+char *image_path = NULL;
+
 /* isutf, u8_dec © 2005 Jeff Bezanson, public domain */
 #define isutf(c) (((c)&0xC0) != 0x80)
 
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for your PR.

Given that i3lock is a security-sensitive program, I would rather not have it reload anything in lock state. Even small mistakes, like a 1-byte memory leak, can manifest themselves as a lock screen circumvention bug.

xPMo:
More features are often accepted in [i3lock-color](https://github.com/Raymo111/i3lock-color).

Issue #164, opened 2017-11-27
Please reinstantiate the red unlock indicator for C-u


xtaran:
Dear Michael,

from the upstream changelog of i3lock 2.10:

 • Immediately hide the unlock indicator after ESC / C-u (Thanks Orestis)

Please revert that (IMHO annoying mis-) feature for at least C-u ~~and the not documented change for Backspace~~.

At least for me the red clearing unlock indicator served for several purpose for me (and probably also for other users):

* To disable the screen blanking without entering any part of a password and still being able to immediately get feedback if the system is still responding to key presses.

* To have some indicator how many Backspace presses have been received by e.g. an heavily loaded system.

I do see a need for a key to quickly hide the unlock indicator, but IMHO the only sense-making (and natural) key for that is the Escape key.

So please revert the above mentioned behaviour for C-u ~~and Backspace~~, but keep it for Esc.

(Initially reported by me [in Debian as #882828](https://bugs.debian.org/882828).)

stapelberg:
> Please revert that (IMHO annoying mis-) feature for at least C-u and the not documented change for Backspace.

I don’t observe any change in backspace, and the code change in question didn’t touch backspace (https://github.com/i3/i3lock/pull/145/files) — what do you mean?

> At least for me the red clearing unlock indicator served for several purpose for me (and probably also for other users):
> 
> To disable the screen blanking without entering any part of a password and still being able to immediately get feedback if the system is still responding to key presses.
> 
> To have some indicator how many Backspace presses have been received by e.g. an heavily loaded system.

You could use the spacebar for the same purpose, or just enter your password. I understand that it doesn’t have exactly the same semantics, but those semantics aren’t a feature we wanted to build into i3lock in the first place.

> I do see a need for a key to quickly hide the unlock indicator, but IMHO the only sense-making (and natural) key for that is the Escape key.

That’s very subjective. i3lock always treated C-u and Escape as synonyms, so I’d like to keep it that way.

xtaran:
Hi Michael,

On Mon, Nov 27, 2017 at 07:33:51AM +0000, Michael Stapelberg wrote:
> > Please revert that (IMHO annoying mis-) feature for at least C-u
> > and the not documented change for Backspace.
> 
> I don’t observe any change in backspace, and the code change in
> question didn’t touch backspace
> (https://github.com/i3/i3lock/pull/145/files) — what do you mean?

Indeed. I just noticed that Backspace still shows the red indicator if
there is something to delete. 

[...]

PR #300, opened 2021-04-11
Changed background color

panos:
Hi, I don't usually try to enforce my own defaults in my open-source contributions, however, as I poked around with the code of `i3lock` I came across the `color` variable.

When I initially tried out i3lock, I remember being woken up by the solid white lock screen color, and it's considerably hard on my own eyes when I'm even near my computer, let alone having to type the password itself. It was a bit tough, and I already even had a headache by the time that I logged in. I've also heard that super bright colors are frustrating to people with sensory issues (I haven't asked before making this change, so I can't be for sure as to whether my own change is helpful or speak for anyone else, but I'm bringing this up since that did happen to be one of the reasons as to why I made that one-line change. Please correct me if I'm wrong.).

I figured that a "pastel" dark color would be considerably better and much less annoying as a default than solid white. I thought of the AMOLED-friendly pitch black color, but, after using it as a default for quite some time, I had to stop, particularly because I had no idea whether my device was actually locked or shut off and I found that particularly frustrating when dealing with laptop lids. I'd be glad to discuss this a bit further and/or change the color.

### DIFF
--- a/i3lock.c
+++ b/i3lock.c
@@ -57,7 +57,7 @@
 typedef void (*ev_callback_t)(EV_P_ ev_timer *w, int revents);
 static void input_done(void);
 
-char color[7] = "ffffff";
+char color[7] = "a3a3a3";
 uint32_t last_resolution[2];
 xcb_window_t win;
 static xcb_cursor_t cursor;


stapelberg:
> I thought of the AMOLED-friendly pitch black color, but, after using it as a default for quite some time, I had to stop, particularly because I had no idea whether my device was actually locked or shut off and I found that particularly frustrating when dealing with laptop lids.

Being able to distinguish a shut-off system from a locked system is precisely the reason why i3lock defaults to white instead of black.

The color you’re suggesting in this pull request is too dark. We need something that definitely is clearly recognizeable as white, even on years-old laptop screens with backlights at the end of their lifecycle.

I know that in web development, there are contrast ratios recommended for accessibility. Are there similar guidelines for what is acceptable in terms of solid colors? I’m curious.

PS: If any color on your screen causes you discomfort, do you have the brightness of your screen too high? I usually find I need to reduce the brightness to 50% or less for my own viewing pleasure.

panos:
> The color you’re suggesting in this pull request is too dark. We need something that definitely is clearly recognizeable as white, even on years-old laptop screens with backlights at the end of their lifecycle.

Acknowledging this.

> I know that in web development, there are contrast ratios recommended for accessibility. Are there similar guidelines for what is acceptable in terms of solid colors? I’m curious.
> 
> PS: If any color on your screen causes you discomfort, do you have the brightness of your screen too high? I usually find I need to reduce the brightness to 50% or less for my own viewing pleasure.

I asked around a bit and did a bit of research, it's the blue light that can generally cause migraines, photophobia, and affect sleeping patterns or the health of the average user negatively. If white's too harmful (LED vs. OLED) and black's too confusing, I guess that we should consider whether there's another middle ground. I'll conduct some further research and write back.

stapelberg:
> LED vs. OLED

FWIW, I’m not concerned with any color’s effect on physical screens.

The lock screen is usually shown briefly when unlocking the computer. If you leave your computer locked for long periods of time, turn off the screen, which will be better for its health, no matter the color.

panos:
> > LED vs. OLED
> 
> FWIW, I’m not concerned with any color’s effect on physical screens.

I'm mostly talking about the effects of blue light and how that said effect can differ depending on the type of the monitor that the user is using, but that's besides the point.

panos:
How about `#464d55` or `#554e46`? They're 67% black on the CMYK scale, but they still hold enough color to pass as a color other than black by default.

From a lighter perspective, I sort of dig `#9d9388` and `#7e7d94`...

stapelberg:
I recently came across a study (https://www.sciencedirect.com/science/article/abs/pii/S2352721821000607) which concludes that blocking blue light might not make any difference, actually:

> Across our full study sample, there were no differences in sleep outcomes attributable to Night Shift.

---

The colors you suggested are far from white. I thought we wanted to stick with white, and maybe tone it down to the minimum acceptable white?

That said, maybe we should re-state what goal we’re trying to achieve. As mentioned previously, LED/OLED friendliness are not of concern, and the above study is counter-evidence to blue light being harmful. Aside from personal preference, which reason remains to change the color at all?

orestisfl:
Michael, blue light might not affect sleep outcomes but there is a plethora of bibliography regarding migraines and other effects: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4939697/

Since this is not our area and we have reports that the default white is uncomfortable for some people, why not change it? What are the arguments in favor of the white light?

stapelberg:
> Since this is not our area and we have reports that the default white is uncomfortable for some people, why not change it?

I’m not opposed to changing it, but I would like to have a good rationale for why we changed it, and what color we changed it to. Currently, we have 1 report of a user who found it uncomfortable, suspected issues with photo-sensitive users, longevity issues with display technology and the assertion that blue light is generally bad for a plethora of reasons, though at least part of that seems to lack good evidence.

Maybe we end up with a rationale of “100% white can be jarring so we toned it down to 75% white”. I just want to avoid justifying changes like that with science, when there isn’t clear science to support our conclusions.

> What are the arguments in favor of the white light?

The current white color is the status quo. Changing it requires a justification, not keeping it :)

panos:
> I’m not opposed to changing it, but I would like to have a good rationale for why we changed it, and what color we changed it to. Currently, we have 1 report of a user who found it uncomfortable, suspected issues with photo-sensitive users, longevity issues with display technology and the assertion that blue light is generally bad for a plethora of reasons, though at least part of that seems to lack good evidence.
> 
> Maybe we end up with a rationale of “100% white can be jarring so we toned it down to 75% white”. I just want to avoid justifying changes like that with science, when there isn’t clear science to support our conclusions.

Did some research but forgot to respond. I got word that people with astigmatism don't do well with white/bright text on fully black backgrounds because my website followed that specific color scheme.

I looked at/skimmed over a couple of papers over the past few weeks that surround the subject, I figured it'd be worthwhile to share them with some of my notes:

- [Effects of blue light on the circadian system and eye physiology](https://www.researchgate.net/publication/303682799_Effects_of_blue_light_on_the_circadian_system_and_eye_physiology) (January 2016) - TL;DR: There has been a transition to environmentally friendly sources of lighting that emit blue light. The abstract makes a sidenote about white light degrades LED screens. Blue light hum, if applied properly, can have an effect in the circadian clock of mice, as well as human beings.
- [European Commission - Scientific Committee on Emerging and Newly Identified Health Risks Section 3.5.1.1](https://ec.europa.eu/health/ph_risk/committees/04_scenihr/docs/scenihr_o_019.pdf) - TL;DR: Small summary describing the correlation between inducing seizures and lamps, depending on their flicker frequency. Contains a lot of additional sources, potentially irrelevant.
- [Effect of Blue Light on Human Eye: Advances to counter its impact (A Review)](https://www.iosrjournals.org/iosr-jdms/papers/Vol19-issue1/Series-3/L1901036266.pdf) (January 2020) - TL;DR; Short review that goes on to correlate the exposure of schoolchildren and adult workers directly to extensive exposure to computer monitors and screens. "[...] it is not surprising that up to 90% of users periodically experience asthenopic symptoms including eyestrain, headaches, ocular discomfort, dry eye, diplopia and blurred vision". Moreover, "it has *not* been studied whether blue light in the displays emitting low intensity light [...] also causes apoptosis".
- [The inner clock—Blue light sets the human rhythm](https://doi.org/10.1002/jbio.201900102) (21 August 2019) - TL;DR: The circadian system is primarily affected the most by blue light. Figure 4 goes in detail about the long-term effects of chronic light exposure.
- [Effects of Dark Mode on Visual Fatigue and Acuity in Optical See-Through Head-Mounted Displays](https://sreal.ucf.edu/wp-content/uploads/2019/10/Kim2019edm.pdf) - TL;DR Interesting experiment that basically concludes that dark backgrounds with white text achieves a higher amount of visual acuity in a selective sample. It was carried out with AR headsets.

panos:
Hm, maybe the whole "I am not sure whether there's a lock screen or whether my computer is suspended" problem could be easily resolved by just rendering the circle from the get-go instead of waiting for keyboard input?

stapelberg:
> Hm, maybe the whole "I am not sure whether there's a lock screen or whether my computer is suspended" problem could be easily resolved by just rendering the circle from the get-go instead of waiting for keyboard input?

Thanks for the suggestion, but there are multiple reasons why I’m not convinced displaying the unlock indicator (“circle”) immediately is a good idea:

1. Users who display an image on their lockscreen will likely not appreciate the unlock indicator being displayed permanently.
2. The unlock indicator can be turned off, and lockscreens should still be recognizeable even when it’s turned off.
3. Users rely on the subtleties of the unlock indicator behavior (see e.g. https://github.com/i3/i3lock/issues/164), so I’m generally cautious in changing anything about it.

---

Thanks also for sharing your research notes. AFAICT, all these papers research either prolongued exposure to certain lights, or performing tasks in certain light environments (e.g. reading text). None of this applies here; we’re rather talking about seeing a flash of light for a few seconds.

I suppose it will be hard to find scientific evidence for the specific, narrow situation at hand.

For now, I’m convinced that using a rationale along the lines of “100% white is perceived as uncomfortable by some users” is better than using any scientific studies.

---

I think the only path forward here is to suggest a color value that is still white, but less than 100% intensity. Do you have a suggestion?

panos:
### Suggestions closer to gray

- `#dadad1`
- `#d9d9d9`
- `#bfbfbf`
- `#b7b7b7`
- `#a3a3a3`

### Suggestions closer to white

- `#f0f0f0`
- `#ededed`
- `#e3e3e3`

orestisfl:
> I think the only path forward here is to suggest a color value that is still white, but less than 100% intensity. Do you have a suggestion?

We can also try `#fdf6e3` which is the solarized light background color.

panos:
> `fdf6e3`

I do have a list of beige-like background colors too, I just wasn't sure whether to include it because it wasn't exactly light. But yeah, that color is not really as blinding and reading text seems to be easier.

stapelberg:
> * a3a3a3

I just tried all suggestions and even this one (the closest to gray) seems fine, i.e. recognizeable as white.

Can you update the commit and commit message accordingly please?

panos:
> Can you update the commit and commit message accordingly please?

Yup, thanks!

stapelberg:
Commit message says a3a3a3, but the commit itself still contains 1d1c1b?

panos:
> Commit message says a3a3a3, but the commit itself still contains 1d1c1b?

Sorry, did this in a rush and failed to add the file, my bad.

stapelberg (approved):

stapelberg:
Thanks!

Raymo111:
Great PR @panos, was going to suggest this myself but never had the courage to.

panos:
> was going to suggest this myself but never had the courage to.

Thanks! It's just a one-line change that was made based on my own experience as a user - and I think you can agree on that it was useful for the same reasons that I also had in the back of my head. I've been making various contributions to @haiku recently and I've been currently spending a lot of time communicating with people and issuing massive changes to its website in order to make other people comfortable, because they're all intimidated by the concept of working on an operating system when it's really just code. It really pains me to see many people with very good ideas end up being intimidated - we all turn out to be losers that way. .-.

We're all in this together! :D (I may not be Dennis Ritchie himself, I'm just a kid that's having a lot of fun with all of this and I've only started contributing over the past year or so for the most part, but do feel more than free to contact me if you need anything revolving around that topic, @Raymo111)

Raymo111:
> It's just a one-line change that was made based on my own experience as a user

UX can make or break projects, and sometimes a one-line contribution is what we need. Have you seen [i3lock-color](https://GitHub.com/Raymo111/i3lock-color)? I've got a large superset of features and a lot to consider for improvements if you'd like to contribute.

panos:
> I've got a large superset of features and a lot to consider for improvements if you'd like to contribute.

I'll give it a check later, thanks! It would be preferable that we don't discuss any further about this on here.

☮️ 🕊️ ✌🏻

PR #317, opened 2022-05-20
Switch build system from autotools to meson

stapelberg:


Airblader (approved):

.gitignore
```diff
@@ -3,29 +3,8 @@
...
-/configure.scan
-/depcomp
-/install-sh
-/missing
-/stamp-h1
+# We recommend building in a subdirectory called build.
+# If you chose a different directory name,
+# it is up to you to arrange for it to be ignored by git,
+# e.g. by listing your directory in .git/info/exclude.
+/build
```
eli-schwartz, line 10: Since Meson 0.57.0, it will automatically create e.g. `builddir/.gitignore` with the contents `*`, so that you don't need to deal with this.

meson.build
```diff
@@ -0,0 +1,144 @@
...
+################################################################################
+
+man1 = join_paths(get_option('mandir'), 'man1')
+
+install_data(
+  [
+	'i3lock.1',
+  ],
+  install_dir: man1,
+)
```
eli-schwartz, line 74: What's wrong with the `install_man('i3lock.1')` function?

meson/meson-dist-script
```diff
@@ -0,0 +1,19 @@
...
+set -eu
+
+cd "${MESON_DIST_ROOT}"
+
+# Delete everything we do not want to have in the release tarballs:
+rm -rf \
+   .clang-format \
+   .editorconfig \
+   travis \
+   .github
```
eli-schwartz, line 12: The primary use case of meson dist is for running on the developer machine, so you may not actually care about supporting dist on really old versions of Meson.

Meson 0.60.0 supports using `.gitattributes` features like `export-ignore` to delete files you don't want in dist tarballs (and also used by github autogenerated archives).

meson/meson-dist-script
```diff
@@ -0,0 +1,19 @@
...
+
+mkdir build
+cd build
+meson .. -Dprefix=/usr
+ninja
+cd ..
+rm -rf build
```
eli-schwartz, line 19: As for this... what is the purpose of this? You're building it but not doing anything with the results, not even copying files out?

stapelberg: Yeah, that was copied from i3. Removing in https://github.com/i3/i3lock/pull/320

meson.build
stapelberg: Thanks, done in https://github.com/i3/i3lock/pull/320

.gitignore
stapelberg: Our minimum meson version is 0.45, which is in Ubuntu 18.04 that’s still supported until 2023. We can probably simplify this eventually, but not just now.

meson/meson-dist-script
stapelberg: Good to know! I’ll leave this as-is for now for consistency with the i3 meson setup. We can probably adopt this once Meson 0.60 is more common. Even though `ninja dist` is primarily run by me, I don’t want to break it just in case it comes in handy for others.

Issue #285, opened 2020-08-08
Allow playing of video on lock (rather than just png/jpg file)


wakatara:
## I'm submitting a…

<pre>
[ ] Bug
[X] Feature Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
i3lock suspends to a blank white screen though you can also sub in a pretty image in there.

## Expected Behavior
It would be pretty schmck if you could lock to a video playing via mpv or such. I'm moving over from OSX and really missing my Aerial lockscreen which plays some amazing undersea vids scenes (It's ridiculously soothing watching the whale and dolphins, I swear.).

## Reproduction Instructions
n/a - though taking a look at  [Aerial](https://github.com/JohnCoates/Aerial) which also has an [xscreensaver port](https://github.com/graysky2/xscreensaver-aerial/) gives you a pretty good idea of what I want to achieve.  =] 
If you have an AppleTV, it's basically what you see when it goes inactive.

## Environment
Output of `i3lock --version`:
<pre>
i3lock version: 2.12 © 2010 Michael Stapelberg
</pre>

Running on Manjaro i3 version 20.0.3 (though pretty heavily customized)

Airblader:
Thanks for the suggestion, but this falls into the eye candy category. We don't really want to trade that complexity into the project as we want to keep i3lock as simple as possible given it's security nature. Thanks for understanding!

wakatara:
@Airblader Totally understand. Could you perhaps point me in the direction of how I might somehow achieve this a la eye candy and simply handing off from i3lock to mpv then? I am just sort of getting into i3lock so have noticed forks and numerous scripts to achieve effects, so pointers would be appreciated (I think I also just stole your compton config from your dotfiles just by coincidence... =p ).

PR #323, opened 2022-06-21
Add GIF support for animated lock screen

sh1r4s3:
Use gif_lib to read a GIF file [1] and show animated lock screen.
Currently it supports only 0-2 disposal methods.

[1] https://www.w3.org/Graphics/GIF/spec-gif89a.txt

demonstation


Airblader:
See #285

PR #324, opened 2022-06-21
unlock_indicator: only show caps lock and num lock (Thanks sandwm)

### LINKED ISSUE #322: Don't (always) show all modifier keys
## I'm submitting a…
<pre>
[x] Bug
[ ] Feature Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
Since v2.14 or more specifically deaf8b9f0f8193b92bdf7f7631f82471296812d1 i3lock shows active modifier keys on each key press. This was implemented in response to issues #316 and #225. However those requests were only about Caps Lock and Num Lock. Yet, i3lock now also shows e.g. Shift being pressed.

## Expected Behavior
i3lock should not show Shift, AltGr and possibly others (depending on keyboard layout) being pressed on the screen. This leaks information about the password if someone can see your screen, e.g. the monitor is visible from behind you, the lock screen is on a beamer because you're about to give a presentation etc.

Possible solutions:

a) \
Only show active modifiers before and after entering a password, i.e. when the password buffer is empty. Caveat: This would still leak the modifier for the first character of the password (e.g. if you begin your password by holding Shift and then pressing P), although it would be much better than doing that for all characters.

b) \
Only show some modifiers. From my perspective, those would be Num Lock and Caps Lock, but going by the comment at https://github.com/i3/i3lock/pull/5#issuecomment-74970333, there might be more to be considered. For Neo (mentioned in the comment), https://neo-layout.org/Benutzerhandbuch/Einrasten/ (German) says, layers 2 and 4 can be locked. I tested Neo and saw that a layer 2 lock is represented via the Lock (aka Caps Lock) modifier. Layer 4 lock shows as Mod2 being active. It's possible that this option might end up at either not consistently working for all lock modifiers on all keyboard layouts (e.g. "sorry, we only support this for Num Lock and Caps Lock"), or encompassing (almost) all modifiers because each one might be a lock somewhere in some keyboard layout...

b2) \
Like b), but from a different starting point: Only exclude specific modifiers. For a regular German keyboard layout, that would be Shift and AltGr – that is Mod5, because that's what AltGr is represented as (while an AltGr modifier still exists, too).

c) \
Add a switch `--show-modifiers={always,empty,never}`, with `empty` implementing a). Or just `--no-modifiers` for a simpler solution.

Considering different keyboard layouts etc., this is getting more complicated than I initially thought. I'm wondering whether other screen lockers just go with Num + Caps Lock or how they handle this. Locally I patched i3lock to do just that (in unlock_indicator.c, line 104ff.), but I think this might be relevant for more people than just me.

## Reproduction Instructions
Run i3lock and enter a password (or other text) where you need modifier keys like Shift for some of the letters / characters.

## Environment
Output of `i3lock --version`:
<pre>
i3lock: version 2.14+ © 2010 Michael Stapelberg
</pre>

stapelberg:
fixes #322


Airblader:
For my understanding, would this handle a situation correctly where someone rebinds say the caps lock key to something else?

stapelberg:
> For my understanding, would this handle a situation correctly where someone rebinds say the caps lock key to something else?

Yes, I think so. I have tested it with the NEO layout, where caps lock is activated by pressing left shift and right shift at the same time, and it did display the caps lock modifier being active.

Issue #287, opened 2020-08-31
Display the keyboard layout on the lock screen.


Liorst4:
## I'm submitting a…

<pre>
[ ] Bug
[x] Feature Request
[ ] Other (Please describe in detail)
</pre>

## Current Behavior
The input language is not displayed on the screen. If you don't remember what was the input language when you locked the machine, you need to switch the language and try until it works.

## Expected Behavior
The input language should be displayed on the screen.

## Reproduction Instructions

```
setxkbmap -option grp:alt_shift_toggle us,es
# Press alt+shift a random amount of times
i3lock
# You have only 50% chance of unlocking the computer if you enter the right password.
```

## Environment
Output of `i3lock --version`:
<pre>
i3lock version: 2.12
</pre>

stapelberg:
We display when you press caps lock, so the keyboard layout falls into the same category.

Unfortunately, things aren’t so easy with keyboard layouts: desktop environments may have the concept of a configured/enabled keyboard layout, but i3 (or other window managers used without desktop environments) don’t have that.

Without such a list, the only thing we could do is display the current keyboard layout *all the time*, which seems excessive to me. Most users only use one keyboard layout, so why would we annoy them with an unhelpful bit of information in a prominent place?

That said, if seeing the keyboard layout at the time of the screen lock on your lock screen, one option you have is to overlay this information on an image yourself in a wrapper script that starts i3lock.

stapelberg:
I’ll tentatively close this as I don’t see a good way of making it happen.

orestisfl:
@Liorst4 you can run `[xkb-switch](https://github.com/grwlf/xkb-switch) -s us` prior to locking to make sure that the correct layout is enabled.

@stapelberg Aren't layouts part of XKB? What do DEs have to do with it?

stapelberg:
> @stapelberg Aren't layouts part of XKB? What do DEs have to do with them?

Layouts are defined via XKB indeed.

What DEs can add is a dialog in which you select which keyboard layouts you actually want to use out of the ≈hundred that are available. This allows for chosing between, say, English and Russian (just 2 entries instead of 100s) in a drop-down (e.g. tray icon).

See the screenshots at http://linuxblog.darkduck.com/2013/10/how-to-configure-keyboard-layouts-in-unity-gnome3-kde.html if you’re unfamiliar.

Liorst4:
```
$ setxkbmap -query
...
layout:     us,es
options:    grp:alt_shift_toggle
```
[...]

PR #341, opened 2023-05-28
Display current keyboard layout when user types password.

vchigrin:
This feature may be very useful for people that have more than one language installed on their systems.

### DIFF
--- a/unlock_indicator.c
+++ b/unlock_indicator.c
@@ -62,6 +62,8 @@ extern char color[7];
 
 /* Whether the failed attempts should be displayed. */
 extern bool show_failed_attempts;
+/* Whether keyboard layout should be displayed. */
+extern bool show_keyboard_layout;
 /* Number of failed unlock attempts. */
 extern int failed_attempts;
 
@@ -331,7 +333,7 @@ void draw_image(xcb_pixmap_t bg_pixmap, uint32_t *resolution) {
             cairo_set_font_size(ctx, 14.0);
             display_button_text(ctx, modifier_string, 28., use_dark_text);
         }
-        if (layout_string != NULL) {
+        if (show_keyboard_layout && layout_string != NULL) {
             cairo_set_font_size(ctx, 14.0);
             display_button_text(ctx, layout_string, -28., use_dark_text);
         }
--- a/unlock_indicator.c
+++ b/unlock_indicator.c
@@ -87,7 +87,7 @@ static xcb_visualtype_t *vistype;
 unlock_state_t unlock_state;
 auth_state_t auth_state;
 
-static void string_append(char** string_ptr, const char* appended) {
+static void string_append(char **string_ptr, const char *appended) {
     char *tmp = NULL;
     if (*string_ptr == NULL) {
         if (asprintf(&tmp, "%s", appended) != -1) {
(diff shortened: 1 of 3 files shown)


unlock_indicator.c
```diff
@@ -85,6 +87,48 @@ static xcb_visualtype_t *vistype;
...
+    if (layout_string) {
+        free(layout_string);
+        layout_string = NULL;
+    }
+    xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(xkb_keymap);
+    for (xkb_layout_index_t i = 0; i < num_layouts; ++i) {
+        if (xkb_state_layout_index_is_active(xkb_state, i, XKB_STATE_LAYOUT_EFFECTIVE)) {
+            const char* name = xkb_keymap_layout_get_name(xkb_keymap, i);
+            if (!name) {
+                name = "Unknown";
```
stapelberg, line 125: It would probably be better to display nothing rather than “Unknown”, as that doesn’t convey to the user what’s missing (but “keyboard layout: unknown” is too long).

unlock_indicator.c
```diff
@@ -280,31 +316,16 @@ void draw_image(xcb_pixmap_t bg_pixmap, uint32_t *resolution) {
...
-            y = BUTTON_CENTER - ((extents.height / 2) + extents.y_bearing) + 28.0;
-
-            cairo_move_to(ctx, x, y);
-            cairo_show_text(ctx, modifier_string);
-            cairo_close_path(ctx);
+            display_button_text(ctx, modifier_string, 28.);
+        }
+        if (layout_string != NULL) {
+            cairo_set_font_size(ctx, 14.0);
+            display_button_text(ctx, layout_string, 56.);
```
stapelberg, line 328: Instead of displaying the layout below the modifier text, can you try displaying it above the `text`?

stapelberg (changes requested):
In the past, we have rejected this feature (see https://github.com/i3/i3lock/issues/287).

But it seems to be a relatively popular and frequent request, and your PR looks reasonably clean, so I’m considering merging it.

The visibility of the layout indicator isn’t great on the dark grey of the unlock indicator. That is a pre-existing issue also affecting the modifier display, but maybe you can try fixing that, too, while you’re at it? Perhaps changing the color to something with more contrast is all we need.

unlock_indicator.c
vchigrin: Done

unlock_indicator.c
vchigrin: done

orestisfl (changes requested):
Nice work, works on my machine

One request: Can you make sure that changing layouts immediately changes the displayed layout in the indicator?

A call to `redraw_screen()` after `process_xkb_event()` here https://github.com/vchigrin/i3lock/blob/be3add9eeed554d16b418f891aaf9df5852e3e74/i3lock.c#L904 works for me, but perhaps this should be done more elegantly.

vchigrin:
Thank you, indeed, putting redraw_screen() there fixes issue with both CapsLock and current keyboard layout display. I've updated this pull request.

stapelberg:
I tried it out and it looks better now.

Before I merge this, can you make displaying the keyboard layout off-by-default, and only visible when a command-line flag was specified? We have a `--show-failed-attempts` flag already, so a consistent choice would be `--show-keyboard-layout`. 

Thanks

stapelberg (approved):

stapelberg:
Thanks for your contribution!

vchigrin:
Thank you for maintaining this project =)

assaron:
@stapelberg Hi, do you plan to make a release with this feature?

stapelberg:
It’s been a while, yeah. I just released i3lock 2.15 with this change.

PR #343, opened 2023-06-22
Adding Space Mode

Joseph-Melberg:
I wake my computer by pressing the space bar, which adds the ' ' to the beginning of my password.

I am tired of this, so I added an option to enable "Space Mode", which prevents the system from taking the space key as a password character until a valid password character is processed.

I have this set to be 'opt-in' in case someone has their password start with ' '.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+2023-06-21 i3lock 2.14.2
+
+ • added Space Mode: mode where space presses aren't
+   registered until a password character is taken.
+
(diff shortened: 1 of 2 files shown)


stapelberg:
Thanks for your pull request, but this doesn’t seem like a feature that’s broadly useful for enough people to justify the maintenance overhead.

I suggest you maintain this as a custom patch, or just wake your computer using a modifier key. E.g., press Shift to wake instead of Space.

PR #362, opened 2024-10-19
add support for more img formats using stb_image.h

pross1312:


stapelberg:
Thanks for your PR, but I’m not interested in making i3lock ship or depend on STB. The warning on the STB website explains that security issues might take considerable time to address, which is not the right level of quality for a screen locker.

See https://github.com/i3/i3lock/issues/68 for prior discussion regarding JPEG support. The most likely candidate seems to be gdk pixbuf.

PR #365, opened 2024-12-13
Update meson setup command in README

eplanet:
This is a very easy fix, when I run the README instructions `meson .. -Dprefix=/usr`, I get the following warning:
```
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```
The fix is to just run `meson setup -Dprefix=/usr` instead and it works like a charm.


stapelberg:
Thanks!

PR #372, opened 2025-09-29
declare a development shell in flake.nix

stapelberg:
For context, I had to declare this file anyway for developing i3lock, so I figured I might as well put it into the repository for the benefit of others. Of course, people who want to ignore Nix are free to keep doing so :) 

From the commit message:

This makes it possible to build i3lock (for development) on any system on which Nix can be installed (= most Linux systems).

For example, I start Emacs using `nix develop --command emacs` and that Emacs process is in an environment with all i3lock build deps.

See also:
https://michael.stapelberg.ch/posts/2025-07-27-dev-shells-with-nix-4-quick-examples/


stapelberg:
Sure, I updated the commit message accordingly.

PR #376, opened 2025-10-28
meson: use explicit_bzero if it is available

julio-b:
It seems that the check for explicit_bzero was lost during the conversion to meson build system.


stapelberg (approved):

stapelberg:
Thanks, good catch! I’ll do a new i3lock release soon.

